PR #8, opened 2014-11-11
get go packages with a single command

toldjuuso:
Hi,

I just noticed that your README instructs people to get each go dependency with separate command. However, implying that the repository is cloned via git, you can get all go dependencies with simply running `go get ./`. on the go application root.

I fiddled with the README already, but did not find a good way to install the dependencies without leaving the application root. Running `go get ./transfersh-server` does not work and using the current patch I made leaves you in the ./transfersh-server folder. Should you have an opinion how to deal with this, I'll tidy my commit as so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,7 @@ $ transfer test.txt
 npm install
 bower install
 
-go get github.com/PuerkitoBio/ghost/handlers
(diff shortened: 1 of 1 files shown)


nl5887:
thx, packages are vendored now.

PR #21, opened 2015-02-25
add t and W

johnko:


johnko:
Also based on a search of SYMBOLS missing the t and W, this repo looks familiar https://github.com/mementor/warper but doesn't have a LICENSE file

nl5887:
Yeah, that repo seems to be inspired at least. Had been nice if there was at least something of a mention. Thanks for the update.

PR #37, opened 2016-02-19
Fix XSS in markdown preview

abpolym:
Markdown can also contain XSS :)


cyberjacob:
@abpolym `EscapeString()` can only be passed a `string`, but the data variable is a `[]byte`. You may want to update your patch to fix this issue, as it is unlikely to be accepted otherwise.

PR #49, opened 2016-05-04
s3: Move from goamz to minio-go

### LINKED ISSUE #5: Implement different storage types than only S3

harshavardhana:
goamz is not maintained anymore move to maintained
library like `minio-go`.

`minio-go` supports S3 and many other s3 compatible
vendors. `minio-go` also supports both Signature V2 and
Signature V4. This in-turn would enable `transfer.sh`
to be able to use many different S3 compatible storage
vendors.

Fixes #5

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,4 @@ Contributions are welcome.
 
 Code and documentation copyright 2011-2014 Remco Verhoef. 
 Code released under [the MIT license](LICENSE). 
+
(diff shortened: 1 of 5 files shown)


zenware:
Hello! This would be awesome for me. I was trying to figure out how to do this earlier today.

harshavardhana:
@nl5887 updated with latest master.

PR #58, opened 2016-07-03
replace white spaces with _ in file name in transfer alias

arastu:
when files have space in their name, transfer alias raise error in curl, this fix error by replacing space with _


nl5887:
Thanks! We've made the sanitization less rigid, you'll be allowed to upload files with spaces and casing now.

PR #30, opened 2015-06-17
add -e to echo in 'transfer' alias example to fix newlines

M1cha:
at least bash needs this.

PR #72, opened 2016-11-27
Fix transfer command

Arinerron:
This pull request fixes the command displayed on transfer.sh index page. It echos the usage, but doesn't enable escape sequences like `\n`, so the guide isn't displayed correctly.

All this pull request does is fix the command by adding `-e` to the command


Arinerron:
Oh, sorry. Didn't see the pull request at https://github.com/dutchcoders/transfer.sh/pull/30

PR #79, opened 2017-03-28
Update README.md alias  for .bashrc

cbanowsky:
added `echo;` to end of alias so that url returned is on single line and easier to read.


aspacca:
@cbanowsky , please, fix conflict if you want the PR merged

PR #84, opened 2017-05-17
Add Max-Hours and Max-Seconds

garex:
Better will be to avoid copy-paste here and to use not headers for params, but url's params, like:
`curl --upload-file ./hello.txt https://transfer.sh/hello.txt?max-downloads=5`

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -319,6 +319,18 @@ func MetadataForRequest(contentType string, r *http.Request) Metadata {
 		metadata.MaxDate = time.Now().Add(time.Hour * 24 * time.Duration(v))
 	}
 
+	if v := r.Header.Get("Max-Hours"); v == "" {
+	} else if v, err := strconv.Atoi(v); err != nil {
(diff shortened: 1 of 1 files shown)


aspacca:
@garex your solution doesn't resolve for multiple expiration headers, but just use an hidden priority (seconds > hours > days), I think the proper solution would be to use `Time.ParseDuration()` on a `Max-Duration` header

garex:
@aspacca I dont' remember about what this PR. If it was not merged about an year, then it's not needed.

aspacca:
@garex , the project was stale, but now I'm helping taking care of it, if you want your still welcome to open a PR with the suggested fix

garex:
I dont' understand your suggestions. Anyway, now I'm stale to this project
))

__________________
Alexander Ustimenko
    +7 (952) 918-02-20

2018-06-17 18:31 GMT+07:00 Andrea Spacca <notifications@github.com>:

PR #89, opened 2017-07-16
Implement support for custom s3 endpoints

kamaln7:
This allows the usage of AWS S3-compatible endpoints such as https://www.minio.io/

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -78,6 +78,12 @@ var globalFlags = []cli.Flag{
 		Value: "",
 	},
 	cli.StringFlag{
+		Name: "s3-endpoint",
+		Usage: "",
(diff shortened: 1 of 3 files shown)


RackerJack:
works for me :)

nl5887:
Cool! Thx.

zenware:
Ultimately this just makes it a function parameter and a command line flag?

kamaln7:
@zenware right. This allows those who want to use a different S3 endpoint to pass it as a config option. The defaults are unchanged.

zenware:
@kamaln7 Awesome, thanks, I'm not super fluent in Go so I appreciate the time you've taken to reinforce my understanding 👍

PR #96, opened 2017-10-03
multi-stages Dockerfile

valentin2105:
Hi,

With these changes, the Docker images move from 269MB to 12.4MB. 

It start FROM a Golang image, and then put the go server in a SCRATCH image.


valentin2105:
It cause issue with CA-Certificates..

PR #97, opened 2017-10-11
Add "Windows support" to README

eteeselink:
Added a section that explains how to get the `transfer` alias on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,19 @@ alias transfer=transfer
 $ transfer test.txt
 ```
 
+### On Windows
+
+Download [curl](https://curl.haxx.se/download.html). Then, put a file called transfer.cmd somewhere in your PATH with this inside it:
+```
+@echo off
+setlocal
+:: write to output to tmpfile because of progress bar
+set tmpfile=%TEMP%\~%~nx1.transfer
+curl --progress-bar --upload-file %1 https://transfer.sh/%~nx1 >> %tmpfile%
+type %tmpfile%
+del %tmpfile%
+```
+
 ## Usage
 
 Parameter | Description | Value | Env
--- a/README.md
+++ b/README.md
@@ -40,11 +40,10 @@ Download [curl](https://curl.haxx.se/download.html). Then, put a file called tra
 ```
 @echo off
 setlocal
-:: write to output to tmpfile because of progress bar
-set tmpfile=%TEMP%\~%~nx1.transfer
-curl --progress-bar --upload-file %1 https://transfer.sh/%~nx1 >> %tmpfile%
-type %tmpfile%
-del %tmpfile%
+:: use env vars to pass names to PS, to avoid escaping issues
+set FN=%~nx1
+set FULL=%1
+powershell -noprofile -command "$(Invoke-Webrequest -Method put -Infile $Env:FULL https://transfer.sh/$Env:FN).Content"
 ```
 
 ## Usage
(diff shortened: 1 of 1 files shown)


eteeselink:
Yes. The first `~` is just a tilde character. Then, `%~nx1` means "the filename and extension of argument 1". So you're correct that it won't work concurrently, _if_ you're uploading multiple files with the exact same filename. 

I bet there's some PowerShell hack to generate a safe random filename like mktemp, but personally I'd think this is good enough. What do you think?

aspacca:
I found this snippet: `[System.IO.Path]::GetTempFileName()`, let me know if it works

eteeselink:
Hmm to be honest I'm not much of a PowerShell guru, but all I do know about is that depending on the Windows version and system security settings, sending stuff from cmd.exe into powershell and somehow assigning the result to a cmd.exe environment variable sounds non-trivial to me (because, I guess, the powershell process would be a child process with its own cloned environment). Also I think that such a solution might easily become significantly more error prone than the current one, which will work for any concurrent uploads except running `transfer hello.txt` in two terminal windows concurrently.

Just to make sure we're on the same page, `transfer hello.txt` in one terminal and `transfer hi.txt` in another will work fine with the current script as well.

aspacca:
> transfer `hello.txt` in one terminal and transfer `hi.txt` in another will work fine with the current script

that's for sure, but I'd rather prefer to avoid putting in the `README` suboptimal information, maybe we could add a wiki page on the project for your solution, with a note about the fact that concurrent multiple upload for files with same base name will not work as expected

eteeselink:
Yes.

But how is that related to this PR? This PR is about getting a `transfer` command that works from anywhere (powershell and cmd). i.e. something that is easy to remember.

Look, I'm just trying to help. I made this PR a long time ago. I believe strongly in a "worse is better" attitude to engineering - i.e. Windows support in 99.9% of cases being better than no Windows support at all (for a human-friendly `transfer` command). At some point, someone else will come by who has a genuis way to do `mktemp` in a cross-cmd/ps, broadly supported way and make a PR. But not if it never gets used of course.

Clearly, you don't want to accept a solution unless it's perfect. Fine too - this one isn't. We disagree fundamentally, and you're the one reviewing this PR, and I'm not. I'll happily help as much as I can but so far I can't see us converging on anything. I don't think calling into PS is a workable solution. I don't think asking people to memorize `Invoke-Webrequest -Method put -Infile .\file.txt https://transfer.sh/file.txt` is just as good as `transfer file.txt`. I think what's in this PR is better than what's out there now. But true, it isn't perfect and I don't know how to make it better.

I won't be offended if you just close this PR. It's OK for people to differ in opinion about this sort of this.

aspacca:
@eteeselink you got me wrong. I was suggesting, in case `invoke-webrequest` is built-in windows, to change your `transfer.cmd` to use it instead of `curl`: we remove a dependencies (that we leave up to the user to find how to satisfy) and we get rid of the problem with the tmpfile.

eteeselink:
Ah! Yes, I like that, actually. I do believe any dev worth their salt has curl somehow, but I see your point.

OK! That's it, you nerd sniped me. Just pushed a new commit. And indeed, this is much better! Turns out that with powershell, you don't even need the tmpfile trick, because the progress bar works differently and is not part of the output. 

Also unless my google-fu has failed me, this *should* work on any modern-ish Windows box, even if all kinds of complicated PowerShell "execution policy" things are set to their (overly restrictive) defaults.

PR #105, opened 2017-11-27
Add: Support for Authenticated Uploads

### LINKED ISSUE #92: Add Support for Authenticated Uploads
In most personal use scenarios, the user wants to lock down upload access to the server. I can't find anything in your `README` that talks about authentication. It is a matter of adding a single `access_token` to each POST request. Any plans for something like this?

Prajjwal:
My stab at implementing authenticated uploads, which is a must have for private deployment scenarios (say, I want to set up file uploads on my own domain, but don't want to open it to the general public). This should resolve #92.

Documented it in the `README` as well.

@dutchcoders, please let me know if you require any changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ listener | port to use for http (:80) | |
 profile-listener | port to use for profiler (:6060)| |
 force-https | redirect to https | false |
 tls-listener | port to use for https (:443) | |
-tls-cert-file | path to tls certificate | | 
+tls-cert-file | path to tls certificate | |
 tls-private-key | path to tls private key | |
 temp-path | path to temp folder | system temp |
 web-path | path to static web files (for development) | |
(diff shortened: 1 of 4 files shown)


Prajjwal:
Is there any interest in merging this?

aspacca:
@Prajjwal your solution doesn't comply with the rfc. The status code to be returned is 401, you should also send back in the response a proper WWW-Authenticate header. Also with custom auth key you won't be able to exploit standard http auth support by the client but build a custom header. If you want to fix this otherwise you're welcome, otherwise I will create my own PR for a proper implementation

Prajjwal:
@aspacca You're right, but I don't see the point of putting any more effort into this. Project appears to be completely dead. If the maintainer ever responds, I'll fix it.

aspacca:
currently in #127

PR #114, opened 2018-02-24
implemented HTTP DELETE for token/filename , local storage provider only

mrizvic:
Hi,

I implemented HTTP DELETE method for token/filename because sometimes I use transfer.sh for sharing large files and want to free up space on server after a couple of successful downloads. DELETE method refuses to delete file (and metafile) if either MaxDays or MaxDownloads arent exceeded. In that case it returns HTTP 500 "MaxDownloads or MaxDays not exceeded yet.".

Since this is the first time I was altering golang code I hope error handling is done right.


mrizvic:
im deleting this repo from my github account

Issue #83, opened 2017-04-26
New awesome exemple (URL copied to clipboard)


heuzef:
Thanks for this amazing tools ! New awesome example with alias (.bashrc) :

`transfer() { if [ $# -eq 0 ]; then echo -e "No arguments specified. Usage:\necho transfer /tmp/test.md\ncat /tmp/test.md | transfer test.md"; return 1; fi
tmpfile=$( mktemp -t transferXXX ); if tty -s; then basefile=$(basename "$1" | sed -e 's/[^a-zA-Z0-9._-]/-/g'); curl --progress-bar --upload-file "$1" "https://transfer.sh/$basefile" >> $tmpfile; else curl --progress-bar --upload-file "-" "https://transfer.sh/$1" >> $tmpfile ; fi; cat $tmpfile | parcellite -c | sleep 1; rm -f $tmpfile; echo "URL copied to clipboard"; echo; }`

URL will be copied to clipboard via [parcellite](parcellite.sourceforge.net) :)

`> transfer texture.jpg 
######################################################################## 100,0%
URL copied to clipboard
`

gingerbeardman:
you should use `pbcopy` instead as it is a built-in macOS command.

Here's my version

```bash
transfer() { if [ $# -eq 0 ]; then echo -e "No arguments specified. Usage:\necho transfer /tmp/test.md\ncat /tmp/test.md | transfer test.md"; return 1; fi
tmpfile=$( mktemp -t transferXXX ); if tty -s; then basefile=$(basename "$1" | sed -e 's/[^a-zA-Z0-9._-]/-/g'); curl --progress-bar --upload-file "$1" "https://transfer.sh/$basefile" >> $tmpfile; else curl --progress-bar --upload-file "-" "https://transfer.sh/$1" >> $tmpfile ; fi; cat $tmpfile | pbcopy | pbpaste; printf '\n'; rm -f $tmpfile; } 
```

PR #115, opened 2018-04-10
Fixed bug with newline character when uploading multiple files

PapiCZ:
This PR fixes bug, when transfer.sh returned links without delimiter `\n` so I've got something like this `https://transfer.sh/1234/file1https://transfer.sh/1234/file2`.

Now it should return something like this:
```
https://transfer.sh/1234/file1(\n)
https://transfer.sh/1234/file2(\n)
```

This bug was created in a3fe3fea. I can't imagine how can this bug exists for more than 1 year :smile: 

Deploy this to production as soon as possible. I want to make app that relies on this :slightly_smiling_face: 

Thank you.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -281,7 +281,7 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 			}
 
 			relativeURL, _ := url.Parse(path.Join(token, filename))
-			fmt.Fprint(w, getURL(r).ResolveReference(relativeURL).String())
+			fmt.Fprint(w, getURL(r).ResolveReference(relativeURL).String() + "\n")
 		}
 	}
 }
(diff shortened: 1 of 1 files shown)


dbgrigsby:
Ran into this issue myself, fixed it in a local use case by adding an echo in the transfer.sh .

gingerbeardman:
This was fixed by several of us in the comments to other issues. The developers never seemed to want to make another update

nl5887:
@gingerbeardman adding a \n will cause issues when copying the link directly to clipboard, that's the reason that \n isn't being added.

gingerbeardman:
My solution worked fine for me in the years I was using it.

YMMV

aspacca:
@gingerbeardman see comment from @kzaidi on #124

gingerbeardman:
>@gingerbeardman see comment from @kzaidi on #124 

@aspacca see mine in #83

PR #128, opened 2018-06-24
ISSUE-44 delete support

### LINKED ISSUE #44: DELETE a file ?
`curl -X DELETE https://transfer.sh/xf54d/myfile.password`
would be awesome.

aspacca:
Fixes #44

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -297,14 +297,17 @@ type Metadata struct {
 	MaxDownloads int
 	// MaxDate contains the max age of the file
 	MaxDate time.Time
+	// DeletionToken contains the token to match against for deletion
+	DeletionToken string
 }
 
 func MetadataForRequest(contentType string, r *http.Request) Metadata {
 	metadata := Metadata{
-		ContentType:  contentType,
-		MaxDate:      time.Now().Add(time.Hour * 24 * 365 * 10),
-		Downloads:    0,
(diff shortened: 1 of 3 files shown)


aspacca:
@nl5887 I cannot test on S3

rogerkerse:
As I understand, transfer.sh has delete support, but how it works? Any examples?

aspacca:
@rogerkerse , @cegohub the upstream version of the source code returns the header `X-Url-Delete` with an url where to send a `DELETE` request to for deleting the file. The header is returned only once on uploading the file. With access to the backend storage is possible to read the metadata to build back the delete url.
Said this, transfer.sh, as hosting service, is not updated to upstream source code, so there's no way to delete a file.

cegohub:
> @rogerkerse , @cegohub the upstream version of the source code returns the header `X-Url-Delete` with an url where to send a `DELETE` request to for deleting the file. The header is returned only once on uploading the file. With access to the backend storage is possible to read the metadata to build back the delete url.
> Said this, transfer.sh, as hosting service, is not updated to upstream source code, so there's no way to delete a file.

Hi, @aspacca  
There is a way to unlist the url path generated from my transfer?

aspacca:
@cegohub what do you mean by unlist? 
anyway I have no control on the hosting service at transfer.sh, you may want to contact directly https://storj.io/

PR #140, opened 2018-07-08
added domain scheme/host for proxied docker containers

gufertum:
Hi,
i added two new parameters to overrule the URL shown after uploading a new file, so 
it also works (shows the correct URL) on a docker container which runs behind a proxy.
Otherwise the URL will be like 'localhost:8080/xyz/test.sh' instead of 'domain.com/xyz/test.sh'.

Let me know, what you think if it....

Regards, Thomas

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 FROM golang:1.7-alpine
-LABEL maintainer="Remco Verhoef <remco@dutchcoders.io>"
+LABEL maintainer="Thomas Schädler <thomas@lambda.li>"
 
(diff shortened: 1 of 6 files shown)


aspacca:
@gufertum I cannot accept this pull request since mainly you're changing package signature (not even everywhere).
also as far as I remember the returned url depends on the requested url, I will investigate further in the next days

gufertum:
Sorry for that, that second part was not intended for the pull request, just for my own local build to test the change - therefore my own package signature.
I will delete this request and file a clean one later on...

Issue #139, opened 2018-07-08
Add domain name as parameter...


gufertum:
I run the docker behind a proxy and the returned path still contains the localhost with the local port.
It would be nice to have a parameter to overrule this when starting the docker container...

gufertum:
I just created a pull request with the changes...
https://github.com/dutchcoders/transfer.sh/pull/140

aspacca:
@gufertum I don't think that adding one or more parameters is the right solution

we should address the problem in server/handlers.go:467 using http header sent by the proxy.
I cannot test the current behaviour at the moment but will do in the next few days

gufertum:
Closed: setting up apache properly, fixed the issue (ProxyPreserveHost On)

PR #141, opened 2018-07-08
Fix hostname when proxied

gufertum:
Ok, this time, only the intended changes are contained in the pull request.
The change allows to use the docker container behind a proxy, by setting the returned domain/url to the correct ("real") value instead of returning 'localhost' (which is correct 'inside') the docker container, but not when proxied to a real domain name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,10 +71,15 @@ gdrive-client-json-filepath | path to client json config for gdrive provider| |
 gdrive-local-config-path | path to local transfer.sh config cache for gdrive provider| |
 lets-encrypt-hosts | hosts to use for lets encrypt certificates (comma seperated) | |
 log | path to log file| | 
+domain-url-scheme | domain url scheme (http or https) when running behind a proxy| |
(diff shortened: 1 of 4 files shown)


aspacca:
@gufertum as I said in the comment of the #139 the proper solution when proxying with a web server is to set the proper host header (ie for nginx: `proxy_set_header Host $host;`)

how do you proxies to docker?

gufertum:
I do the same with apache
`ProxyPass / http://localhost:10000/
  ProxyPassReverse / http://localhost:10000/`.
My fix is just a workaround to get it working, so using the correct headers would be nice, but I'm not very good at go :-)

aspacca:
@gufertum on apache you should use `ProxyPreserveHost On`
https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypreservehost

gufertum:
Thanks for the hint 👍 After fixing my apache configuration, it's working properly.
No more clunky workaround needed...

Issue #148, opened 2018-08-12
No documentation for creating direct download links for files


limitedeternity:
But still in `/server/server.go` following lines exist:

```go
r.HandleFunc("/get/{token}/{filename}", getHandlerFn).Methods("GET")
r.HandleFunc("/download/{token}/{filename}", getHandlerFn).Methods("GET")
```

Which send file without showing a preview page.

Issue #147, opened 2018-08-12
HEAD requests to file result in 405


limitedeternity:
Because `/server/server.go` has no defined handlers for this method. Only GET method allowed:

```go
r.HandleFunc("/{token}/{filename}", getHandlerFn).Methods("GET")
r.HandleFunc("/get/{token}/{filename}", getHandlerFn).Methods("GET")
r.HandleFunc("/download/{token}/{filename}", getHandlerFn).Methods("GET")
```

Otherwise, I see no way to check if file actually exists without GETting it.

PR #149, opened 2018-08-12
Fixes

limitedeternity:
Resolves [#147](https://github.com/dutchcoders/transfer.sh/issues/147) & [#148](https://github.com/dutchcoders/transfer.sh/issues/148)

Adds embedding functionality using `content-disposition: inline; filename=`

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -791,6 +791,33 @@ func (s *Server) tarHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *Server) headHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+
+	token := vars["token"]
+	filename := vars["filename"]
+
+	if err := s.CheckMetadata(token, filename); err != nil {
+		log.Printf("Error metadata: %s", err.Error())
+		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+		return
+	}
+
+	contentType, contentLength, err := s.storage.Head(token, filename)
+	if s.storage.IsNotExist(err) {
+		http.Error(w, http.StatusText(http.StatusNotFound), http.StatusNotFound)
+		return
+	} else if err != nil {
+		log.Printf("%s", err.Error())
+		http.Error(w, "Could not retrieve file.", 500)
+		return
+	}
+
+	w.Header().Set("Content-Type", contentType)
+	w.Header().Set("Content-Length", strconv.FormatUint(contentLength, 10))
+	w.Header().Set("Connection", "close")
+}
+
 func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -818,45 +818,10 @@ func (s *Server) headHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Connection", "close")
 }
 
-func (s *Server) embedHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


limitedeternity:
1. Is `/get/...` intended to be an alias for `/download/...`?
If so, I'll add this line back.

2. Yep, I agree that `inline` is more suitable here than `embed`. 

3. Creating `Link aliases` section seems good idea for me.

4. `HEAD` requests are useful for "pinging" resources without actually getting them. Without `HEAD` support we need to send `GET` request which results in whole 1GB+ file (for example) downloaded. That's quite unwanted behavior if you only need to check response's status code (for instance, to check if remote file still exists).

5. Duplicating `getHandler` isn't good approach indeed. Will change that soon.

aspacca:
4. I was probably not so clear about HEAD request support: I find it very useful and it's welcomed. I'm not sure if adding HEAD support to every urls or only the default one

limitedeternity:
Like that. `Action` variable covers `download`, `get`, `inline`, `retrieve` etc. This allows us to change `content-disposition` in `getHandler` accordingly.

Also made "Link aliases" section in readme.

server/server.go
```diff
@@ -324,6 +324,9 @@ func (s *Server) Run() {
 	r.HandleFunc("/({files:.*}).tar", s.tarHandler).Methods("GET")
 	r.HandleFunc("/({files:.*}).tar.gz", s.tarGzHandler).Methods("GET")
 
+	r.HandleFunc("/{token}/{filename}", s.headHandler).Methods("HEAD")
+	r.HandleFunc("/{action}/{token}/{filename}", s.headHandler).Methods("HEAD")
```
aspacca: please. change to `/{action:(download|get|inline)}/{token}/{filename}`
I'd rather prefer to limit the aliases

server/server.go
```diff
@@ -352,8 +355,7 @@ func (s *Server) Run() {
 	}
 
 	r.HandleFunc("/{token}/{filename}", getHandlerFn).Methods("GET")
-	r.HandleFunc("/get/{token}/{filename}", getHandlerFn).Methods("GET")
-	r.HandleFunc("/download/{token}/{filename}", getHandlerFn).Methods("GET")
+	r.HandleFunc("/{action}/{token}/{filename}", getHandlerFn).Methods("GET")
```
aspacca: same as above

server/handlers.go
```diff
@@ -817,8 +845,13 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", contentType)
 	w.Header().Set("Content-Length", strconv.FormatUint(contentLength, 10))
-	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", filename))
-	w.Header().Set("Connection", "close")
+	w.Header().Set("Connection", "keep-alive")
+
+	if action == "inline" {
```
aspacca: a little redundant. why not setting a `disposition` variable having either `inline` or `attachment` as value and format it in a single `fmt.Sprintf("%s; filename=\"%s\"", disposition, filename)`?

limitedeternity:
Should be alright now.

limitedeternity:
Oh, well, hold on...

limitedeternity:
Well, it's not me. Seems to be a build timeout.

aspacca:
please, push another commit so to restart the build

aspacca:
also it must be: `/{action:(?:download|get|inline)}/{token}/{filename}` not `/{action:(?:(download|get|inline))}/{token}/{filename}`

limitedeternity:
Ah, `Only non-capturing groups are accepted`...

limitedeternity:
Done!

PR #154, opened 2018-09-01
Updated 'Upload' section to insert '-X PUT' in the command 

angsuman:
Updated 'Upload' section to insert '-X PUT' in the command as the default POST request for file upload is failing at this time.


aspacca:
@angsuman I cannot reproduce the problem, also I'm not so keen to update the documentation for a transient problem

can you give more details about your issue? like curl version, full request and response headers etc

PR #155, opened 2018-09-02
Fix post handler

N6UDP:
it would appear that when the authhandler was added the posthandler was replaced with the puthandler breaking post requests.


aspacca:
you're right @N6UDP , thank you very much! will merge as soon a travis is green

PR #156, opened 2018-09-02
Add newline between URLs for multi-upload

N6UDP:
Sorry, also just noticed that you get the response urls concatenated with no separator when uploading multiple files.  That said this will also add a newline character that isn't currently there for POST only requests with one file.. it would be reasonable to avoid adding it on the last or always adding a newline event for PUT requests responses to be consistent.  What do you think? 
Thanks for the quick response on the other PR.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -300,7 +300,7 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 			}
 
 			relativeURL, _ := url.Parse(path.Join(token, filename))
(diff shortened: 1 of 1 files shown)


aspacca:
@N6UDP good catch. anyway @nl5887 was categoric on keeping the actual behaviour of not appending a newline on the url, in order to not break compatibility. I agree with him, so please, refactor your PR to avoid adding it on the last

N6UDP:
@aspacca done; haven't done almost any Go before, so welcome suggestions to make it cleaner.

PR #162, opened 2018-10-06
gpg encryption support

francescayeye:
`X-Encrypt-Password`/`X-Decrypt-Password` headers work seemingless  as in gpg piping shell examples


aspacca:
- missing tests
- missing updated doc

JacksonChen666:
criticism: does not work as end-to-end encryption since the server is definitely receiving the contents unencrypted even under https. also, this feature might actually cause unnecessary overhead (encryption and decryption), requires trust in the provider (there's a reason why you might not trust any providers), and could be useless anyways if the server refuses to give the encrypted version if no password was given.

aspacca:
@stefanbenten what do you think of this?
I've refectored in order to use a maintained opengpg library

stefanbenten:
I'll review this today :)

aspacca:
@stefanbenten did you have a chance to review?
I'd like to merge it

server/handlers.go
```diff
@@ -54,6 +54,10 @@ import (
 	textTemplate "text/template"
 	"time"
 
+	"github.com/ProtonMail/go-crypto/openpgp"
+	"github.com/ProtonMail/go-crypto/openpgp/armor"
+	"github.com/ProtonMail/go-crypto/openpgp/packet"
+	"github.com/ProtonMail/gopenpgp/v2/constants"
```
stefanbenten, line 60: NIT: this is not properly sorted with the below list.

server/handlers.go
```diff
@@ -90,6 +94,124 @@ func initHTMLTemplates() *htmlTemplate.Template {
...
+
+	config := &packet.Config{
+		DefaultCipher: packet.CipherAES256,
+	}
+
+	var emptyKeyRing openpgp.EntityList
+	md, err := openpgp.ReadMessage(unarmored.Body, emptyKeyRing, prompt, config)
+	if err != nil {
+		// Parsing errors when reading the message are most likely caused by incorrect password, but we cannot know for sure
+		return
```
stefanbenten, line 138: I like to be specific about returning with the err like `return err`, but that is personal preference.

stefanbenten (approved):
Small code styling comments, overall looks good.
Will do a staging test now, if you do not hear from me until midnight UTC, assume it worked :)

issadarkthing:
Has this been tested yet?

server/handlers.go
aspacca: it seems that uppercase is order before lower one

server/handlers.go
aspacca: it's this way everywhere in the code, let's open a new PR for this

aspacca:
> Has this been tested yet?

going to merge soon

aspacca:
indeed will wait https://github.com/dutchcoders/transfer.sh/pull/527 before

PR #201, opened 2019-03-19
Add second S3 Handler for Non-Multipart Uploads

stefanbenten:
With this small change, the service is able to upload to custom S3 Endpoints that dont fully support Multipart Uploads.

I added a flag to the runtime, to disable multipart uploads (which is the default and preferred way)

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -2,15 +2,13 @@ package cmd
 
 import (
 	"fmt"
-
+	"log"
 	"os"
-
 	"strings"
 
 	"github.com/dutchcoders/transfer.sh/server"
 	"github.com/fatih/color"
 	"github.com/minio/cli"
-	"log"
 	"google.golang.org/api/googleapi"
 )
 
@@ -116,6 +114,10 @@ var globalFlags = []cli.Flag{
 		Usage:  "",
 		Value:  "",
 		EnvVar: "BUCKET",
+	},
+	cli.BoolFlag{
+		Name:  "no-multipart",
+		Usage: "Disables Multipart Puts",
 	},
 	cli.StringFlag{
 		Name:  "gdrive-client-json-filepath",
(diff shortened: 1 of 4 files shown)


server/storage.go
```diff
@@ -202,7 +201,7 @@ func (s *S3Storage) Delete(token string, filename string) (err error) {
 	return
 }
 
-func (s *S3Storage) Put(token string, filename string, reader io.Reader, contentType string, contentLength uint64) (err error) {
+func (s *S3Storage) PutMulti(token string, filename string, reader io.Reader, contentType string, contentLength uint64) (err error) {
```
aspacca: please, don't export this method as it is not part of the `Storage` interface

server/storage.go
```diff
@@ -316,6 +315,25 @@ func (s *S3Storage) Put(token string, filename string, reader io.Reader, content
 	return
 }
 
+func (s *S3Storage) Put(token string, filename string, reader io.Reader, contentType string, contentLength uint64) (err error) {
+	key := fmt.Sprintf("%s/%s", token, filename)
```
aspacca: you have now `key` defined two times: here and at https://github.com/dutchcoders/transfer.sh/pull/201/files#diff-8cf047e4c4eb7b64ea58583ff04a3110R205

please, remove the second and refactor `PutMulti` method to accept directly a key param

cmd/cmd.go
```diff
@@ -117,6 +115,10 @@ var globalFlags = []cli.Flag{
 		Value:  "",
 		EnvVar: "BUCKET",
 	},
+	cli.BoolFlag{
+		Name:  "no-multipart",
```
aspacca: please, rename to `s3-no-multipart`
also remember to add documentation in `README.md`

server/storage.go
stefanbenten: Done

cmd/cmd.go
stefanbenten: Done

server/storage.go
stefanbenten: Refactored 👍

server/storage.go
```diff
@@ -125,17 +124,18 @@ func (s *LocalStorage) Put(token string, filename string, reader io.Reader, cont
...
 	Storage
-	bucket *s3.Bucket
-	logger *log.Logger
+	bucket      *s3.Bucket
+	logger      *log.Logger
+	noMultipart bool
 }
 
-func NewS3Storage(accessKey, secretKey, bucketName, endpoint string, logger *log.Logger) (*S3Storage, error) {
+func NewS3Storage(accessKey, secretKey, bucketName, endpoint string, logger *log.Logger, multipart bool) (*S3Storage, error) {
```
aspacca: please, rename to `disableMultipart`

last thing then I can merge

stefanbenten: Done

stefanbenten:
@aspacca Finally got to pushing this small fix.
Thanks for taking a look that quickly!

Issue #203, opened 2019-03-28
Support for customized subroute?


lucernae:
Hi! Love this work, @dutchcoders !
Transfer.sh is very easy to deploy and try.

I have some questions on this.
We are planning to use transfer.sh in our web apps as a subroute, in a sense like this:

Let's say I deployed a web app with domain: `domain.org`, and accessible in this url `http://domain.org/`.
I would like to deploy transfer.sh in a subroute (because the web apps is a separate system), like this:
`http://domain.org/transfer` to acces transfer.sh.

I tried to reroute the traffic using haproxy, but it seems the html inside is bound to the route url. So to summarize, I have broken links for scripts, styles, etc, because it gives links to `http://domain.org/scripts` instead of `http://domain.org/transfer/scripts`.

So, my question is, does this use case currently handled by transfer.sh?
If not, should I make PR to do that (I'm not too familiar with golang, so it might take time)?

aspacca:
hello @lucernae you're right
please test #206 branch and tell me if it fixes the problem
I tested only on UI side and it works

please, not that a new param --proxy-path is needed to instruct transfer.sh to build the proper redirect url

lucernae:
Thanks, for the quick follow up! 
I ended up doing some modification (of course it's a hack :p), to make transfer.sh web shows the correct url and all the styles.
I'm not too confident with the code anyway, so I will test your branch code.
It's interesting to see the problem in issue #204  too since it was dealing with https terminated in the proxy.
I will check your branch with the X-Forwarded-Proto header set for https.

lucernae:
I did some quick testing on my part:
Let's say, I deployed on `http://domain.org/transfer` using a combination of haproxy docker container and transfer.sh (go binary).

docker-compose.yml:

```
version: '2'
services:
  transfer:
    image: haproxy:latest
    volumes:
      - ./haproxy:/usr/local/etc/haproxy:ro
    ports:
      - "80:80"
```

haproxy.cfg:
I test using http mode for now. Because there are still some issues.

```
global
daemon

frontend 80
bind *:80
mode http
acl transfer_path path_sub -i /transfer
use_backend transfer if transfer_path

backend transfer
acl forwarded_proto hdr_cnt(X-Forwarded-Proto) eq 0
acl forwarded_port hdr_cnt(X-Forwarded-Port) eq 0
http-request add-header X-Forwarded-Port %[dst_port] if forwarded_port
mode http
[...]

Issue #204, opened 2019-03-29
Lack of compatibility with reverse proxy


Akito13:
I set up an nginx vhost to proxy the requests to the local transfer.sh instance. Therefore, I have to designate localhost:port as the listener for requests to transfer.sh. (Btw. the documentation is horribly wrong, one aspect is that you actually have to provide the _full address_ not only the port, when specifying port arguments.)
The problem now is that when I upload a file, it responds with a link containing the `localhost:port/wa3tr/test.txt` URL instead of the one my reverse proxy listens to. I don't see a way to tell transfer.sh, that it should output a different URL, even if it listens on something entirely different.

All I need is the possibility to override the default output after a file has been uploaded.

aspacca:
hello @Akito13 , you're right
please test #206 branch and tell me if it fixes the problem
I tested only on UI side and it works

aspacca:
please, not that a new param `--proxy-path` is needed to instruct transfer.sh to build the proper redirect url

aspacca:
sorry @Akito13 , wrong issue

as for the proxy support, there are actually two problem: one is the path and one is the host
the path issue in solved in #206 (please, note that a new param --proxy-path is needed to instruct transfer.sh to build the proper redirect url), I ask you to test that branch

as for the host issue: you have to instruct your proxy to rewrite the host header with the original one, instead of overwriting with the one to the proxied service
on nginx you can achieve this with `proxy_set_header Host $host;`

Akito13:
@aspacca First of all, thank you for your quick response and action.

I checked out to ISSUE-203 but I get
```
flag provided but not defined: -proxy-path
```
when providing
```
--proxy-path domain.dom
```
It removes the first dash.

aspacca:
—proxy-path should be a path, not a domain 
so if you are proxing from http://yourdomin.com/apath to transfer.sh bound on localhost:8080 you should pass —proxy-path apath and set your proxy to erite as host header yourdomain.com

I may have misspelled the number of dashes required for the param to be accepted

Akito13:
> 
> —proxy-path should be a path, not a domain so if you are proxing from http://yourdomin.com/apath to transfer.sh bound on localhost:8080 you should pass —proxy-path apath and set your proxy to erite as host header yourdomain.com I may have misspelled the number of dashes required for the param to be accepted
> […](#)
[...]

PR #206, opened 2019-03-30
Issue 203

aspacca:
solves #203


mckaygerhard:
hi! i wish to know if the new options/parameter will need apsolute paths/routes or can allow relatives ? that are very important and seems are more need to be/allow relatives to really solves #203 and #204

PR #214, opened 2019-04-24
Fix syntax coloring for fish functions

maxnordlund:
I guess they were copy-pasted and forgot to change the `bash` to `fish`.


aspacca:
thanks @maxnordlund

maxnordlund:
No problem :)

Issue #177, opened 2018-11-26
Document supported headers on transfer.sh


marcelstoer:
One example on https://transfer.sh/ uses `-H "Max-Downloads: 1" -H "Max-Days: 5"`. I can guess what those headers do but it would be nice if they were properly documented somewhere. Suggestion: README.md.

aspacca:
@marcelstoer yes, I planned to add the full documentation of the supported headers in request and response in the README.md but I haven't had time yet. Feel free to open a PR for it

marcelstoer:
> Feel free to open a PR for it

I would if I knew the answers. However, since I would have to dig into a code base of a language I'm not familiar with this wouldn't be effective at all.

aspacca:
fixed with #225

PR #225, opened 2019-05-11
Document supported headers (#177)

dopessoa:
Documenting response and request headers as requested on issue #177.


aspacca:
@dopessoa thank you!

PR #227, opened 2019-05-15
Fix the bash functions to handle files with spaces

sente:
Minor tweak to the example `transfer` bash functions to handle files with spaces.


aspacca:
thanks @sente

aspacca:
@sente it seems that on the frontend part the quotes were already present, do you mind to double check if you have time?
https://github.com/dutchcoders/transfer.sh-web

thanks

Issue #219, opened 2019-04-28
Update Dockerfile from n8225/transfer.sh


aspacca:
@n8225, you edited Dockerfile to build static and use scratch image for a smaller container

I would like to include the change in the official repository, but it would be nice if you open the PR for it

dopessoa:
As @n8225 hasn't answered in quite a while I've made some improvements to the Dockerfile based on his changes, I've sent a PR with the new Dockerfile.

aspacca:
fixed by #231

PR #231, opened 2019-06-04
Update Dockerfile generating smaller image

dopessoa:
Generate a statically linked binary, somewhat deal with issue #219.

As @n8225 didn't answer in more than a month I've improved the Dockerfile based on his changes.

The resulting image is considerably smaller:

```
dopessoa@local transfer.sh (master) $ docker images | grep transfer.sh
local/transfer.sh         latest              13186b203d06        9 minutes ago       15.4MB
dutchcoders/transfer.sh   latest              1ff531329501        4 months ago        325MB
```

Just bear in mind that the intermediary image get's to near 870MB.

An addition is that you can pass the Golang version of the build image using as a command line argument:

`docker build --build-arg GO_VERSION=1.11 .`

If no argument is present the default is 1.12.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,23 @@
-FROM golang:1.12-alpine as build
-LABEL maintainer="Andrea Spacca <andrea.spacca@gmail.com>"
+# Default to Go 1.12
+ARG GO_VERSION=1.12
+FROM golang:${GO_VERSION}-alpine as build
 
-# Copy the local package files to the container's workspace.
+# Necessary to run 'go get' and to compile the linked binary
+RUN apk add git musl-dev
+
 ADD . /go/src/github.com/dutchcoders/transfer.sh
 
-# build & install server
-RUN go build -o /go/bin/transfersh github.com/dutchcoders/transfer.sh
+WORKDIR /go/src/github.com/dutchcoders/transfer.sh
 
-FROM golang:1.11-alpine
-COPY --from=build /go/bin/transfersh /go/bin/transfersh
+ENV GO111MODULE=on
+
+# build & install server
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,14 +1,23 @@
...
-# Copy the local package files to the container's workspace.
 ADD . /go/src/github.com/dutchcoders/transfer.sh
 
+WORKDIR /go/src/github.com/dutchcoders/transfer.sh
+
+ENV GO111MODULE=on
+
 # build & install server
-RUN go build -o /go/bin/transfersh github.com/dutchcoders/transfer.sh
+RUN go get -u ./... && CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags -a -tags netgo -ldflags '-w -extldflags "-static"' -o /go/bin/transfersh github.com/dutchcoders/transfer.sh
```
aspacca: you don't need `go get` if you set `GO111MODULE=on`: dependency will be automatically fetched when building

Dockerfile
```diff
@@ -1,14 +1,23 @@
...
+
+FROM scratch AS final
+LABEL maintainer="Andrea Spacca <andrea.spacca@gmail.com>"
 
-FROM golang:1.11-alpine
-COPY --from=build /go/bin/transfersh /go/bin/transfersh
+COPY --from=build  /go/bin/transfersh /go/bin/transfersh
 
 ENTRYPOINT ["/go/bin/transfersh", "--listener", ":8080"]
```
aspacca: we can think about removing `"--listener", ":8080"` for allowing people to override this value. but this will break current setup using the image

aspacca:
I've little time lately, thanks @dopessoa  for contributing to the project

the main problem is that the docker hub automatic building is broken since few months and I cannot fix the problem without @nl5887

Issue #229, opened 2019-05-22
Append Max-Downloads and Max-Days on response headers


dopessoa:
Are there any plans to add Max-Downloads and Max-Days current values as response headers on GET and HEAD requests?

I guess this is slightly related to #178.

aspacca:
now there is, I don't know when I'll have time to do it anyway

aspacca:
but I guess the headers should be Remaining-Downloads and Remaining-Days to be useful

dopessoa:
OK, I've been really busy lately, but I'll send a PR in the next few days.

aspacca:
really busy here as well, if you plan to send a PR for this I will wait for it

dopessoa:
Sent a PR, please review when you got the time.

Found what I think is an issue while developing this, I'll open a new issue to document my finding.

PR #234, opened 2019-06-09
Add X-Remaining-Downloads and X-Remaining-Days headers to HEAD and GET responses

dopessoa:
This commit satisfies the request on issue #229.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -564,6 +564,14 @@ func getURL(r *http.Request) *url.URL {
 	return u
 }
 
+func calcRemainingLimits(metadata Metadata) (int, int) {
+	remainingDownloads := metadata.MaxDownloads - metadata.Downloads
+	timeDifference := metadata.MaxDate.Sub(time.Now())
+	remainingDays := int(timeDifference.Hours()/24) + 1
+
+	return remainingDownloads, remainingDays
+}
+
 func (s *Server) Lock(token, filename string) error {
 	key := path.Join(token, filename)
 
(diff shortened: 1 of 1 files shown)


server/handlers.go
```diff
@@ -583,6 +583,35 @@ func (s *Server) Unlock(token, filename string) error {
 	return nil
 }
 
+func (s *Server) GetMetadata(token, filename string) (Metadata, error) {
```
aspacca: please, don't add a 99% identical method (see `CheckMetadata`), just refactor the existing one to return `Metadata`

also related to head requests increasing the download count `CheckMetadata` can become:
`func (s *Server) ProcessMetadata(token, filename string, increaseDownload bool) (Metadata, error) {`

server/handlers.go
```diff
@@ -870,6 +899,8 @@ func (s *Server) headHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	metadata, err := s.GetMetadata(token, filename)
```
aspacca: see first comment: we are doing basically two times the same thing in https://github.com/dutchcoders/transfer.sh/pull/234/files#diff-105fedf355a9fcad82982745cef24f7aR896 and here

server/handlers.go
```diff
@@ -918,10 +957,16 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 		disposition = "attachment"
 	}
 
+	remainingDownloads := metadata.MaxDownloads - metadata.Downloads
```
aspacca: please, extract these three repeated lines from here https://github.com/dutchcoders/transfer.sh/pull/234/files#diff-105fedf355a9fcad82982745cef24f7aR914 in separate methods of `MetaData` encapsulating the logic

dopessoa:
Thank you for the review, I've made some changes. It's much better now.

This PR fixes #235 as well. Should I separate the commits?

aspacca:
thanks @dopessoa , looks good to me. no need to create two separate PR
can you check the failure in travis?

Issue #215, opened 2019-04-26
Percent-encoding in filenames not handled correctly


JustAnotherArchivist:
Percent-encoding in filenames is not handled correctly:

    > echo foo > test%C3%A4test
    > curl --upload-file test%C3%A4test https://{redacted}/;echo
    https://{redacted}/LN3I1/test%C3%A4test

Loading that URL produces a 404 because the file was actually uploaded as `https://{redacted}/LN3l1/test%25C3%25A4test` instead, i.e. with the per cent symbols encoded as `%25`. That's probably the correct way to handle such uploads, but the output of the PUT request should also reflect it.

aspacca:
you're right @JustAnotherArchivist : the current solution was a little clumsy. I fixed with a more straightforward one in #216 , can you give a try?

JustAnotherArchivist:
Thanks, I can confirm that #216 fixes this.

PR #236, opened 2019-06-13
Fix encoding of spaces in filenames from + to %20

JustAnotherArchivist:
Follow-up to #215/#216

Spaces in filenames are incorrectly encoded as `+` instead of `%20`:

    echo foo > 'a file'
    curl -v --upload-file 'a file' https://redacted/
    ...
    < X-Url-Delete: https://redacted/dcr2r/a+file/1pz0rU0YFu
    ...
    https://redacted/dcr2r/a+file

These URLs are 404s. The correct URLs are `https://redacted/dcr2r/a%20file/1pz0rU0YFu` and `https://redacted/dcr2r/a%20file`, respectively.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -328,7 +328,7 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 
 			}
 
-			filename = url.QueryEscape(filename)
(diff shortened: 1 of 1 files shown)


aspacca:
@JustAnotherArchivist thanks

Issue #242, opened 2019-06-21
Support for AWS Signature Version 4


computeralex92:
It would be great if a new version of transfer.sh support the AWS Signature Version 4 to host the S3 Bucket in a region where only this signature is available.

Currently this error appears when trying to upload a file to such a S3 Bucket:
```
[transfer.sh]2019/06/21 06:07:31 The authorization mechanism you have provided is not supported. Please use AWS4-HMAC-SHA256.
```

aspacca:
@computeralex92 can you point to some documentation about the topic?
the S3 storage backend was done by @nl5887 and I never worked on it (neither I have the possibility to test against it at the moment).
I can check if I can just override the AWS Signature to use from the AWS SDK, is Version 4 supported in every region?

computeralex92:
The issue is related to the switch from Version 2 of the signature to Version 4 happen step by step from AWS.
Every newer region (e.g. eu-central-1 or ap-east-1) is only accessable with version 4, all other "older" regions of AWS supports both versions.

As far as I can see in the code the problem is that the S3 Lib transfer.sh is using does not support V4 and since the lib is unmaintained, from my perspective only a switch of the lib is solving this issue.

I found a old pull request which was never merged to master, maybe it is helpful in this case:
https://github.com/dutchcoders/transfer.sh/pull/49

aspacca:
I see @computeralex92. as I said I don't use S3 and I didn't developed that part.
I think the whole code has to be moved to use official AWS SDK, unlikely I don't know when I'll have time for that and I cannot test.
in case I will create a branch for are you available to test it?

aspacca:
please @computeralex92 can you test #243 ?

computeralex92:
> please @computeralex92 can you test #243 ?

I will try it and give you a result as soon as possible.

computeralex92:
OK, looks good so far, but I need to fix some issues to get it work:

- A S3 Bucket with SSL Support is not accessable since the CA Certificates are not copied to the docker image
- The region is hard coded to "eu-west-1", so every other region is failing.
I already fixed this bugs locally and it works now with a bucket located in Frankfurt (eu-central-1).

To things are on my ToDo list for now:

- [ ] Test the new code against Minio and some other S3 compatible providers
- [ ] Find out why the download page of a file shows 0 KB as a file size.

computeralex92:
Good message: It is working as expected with Minio and other providers like Digitalocean and Scaleway.
Bad message:
[...]

PR #244, opened 2019-06-23
Fix SSL and make change of region possible

computeralex92:
To address some issues in #243 from issue #242, I made the following changes:

- Add the SSL Certs to the last docker image
- Make the S3 region changeable via a environment variable & a CLI option

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -99,7 +99,7 @@ var globalFlags = []cli.Flag{
 	cli.StringFlag{
 		Name:   "s3-endpoint",
 		Usage:  "",
-		Value:  "http://s3-eu-west-1.amazonaws.com",
+		Value:  "",
 		EnvVar: "S3_ENDPOINT",
 	},
 	cli.StringFlag{
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -103,6 +103,12 @@ var globalFlags = []cli.Flag{
 		EnvVar: "S3_ENDPOINT",
 	},
 	cli.StringFlag{
+		Name:   "s3-region",
+		Usage:  "",
+		Value:  "eu-west-1",
+		EnvVar: "S3_REGION",
+	},
+	cli.StringFlag{
 		Name:   "aws-access-key",
 		Usage:  "",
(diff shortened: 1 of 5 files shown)


cmd/cmd.go
```diff
@@ -332,7 +338,7 @@ func New() *Cmd {
 				panic("secret-key not set.")
 			} else if bucket := c.String("bucket"); bucket == "" {
 				panic("bucket not set.")
-			} else if storage, err := server.NewS3Storage(accessKey, secretKey, bucket, c.String("s3-endpoint"), logger, c.Bool("s3-no-multipart")); err != nil {
+			} else if storage, err := server.NewS3Storage(accessKey, secretKey, bucket, c.String("s3-endpoint"), logger, c.Bool("s3-no-multipart"), c.String("s3-region")); err != nil {
```
aspacca: note that here `s3-region` can be empty
I would do something like:
```
			region := "eu-west-1"
			if awsRegion := c.String("s3-region"); awsRegion != "" {
				region = awsRegion
			}
```

I set default to `eu-west-1` to not break current installation

the same empty value problem can be for `s3-endpoint`: do you mind testing setting the region but not the endpoint and see if the sdk is smart enough to figure the second?

computeralex92: The SDK is so smart ;-)
If you remove the endpoint and only set the region, it automatically set the correct endpoint in the background.
I would suggest to only remove the default endpoint and let the region default set to eu-west-1.

In the case someone is using e.g. Scaleway they must set region and endpoint to get a success connection to the object storage.

aspacca: so let's go for setting default region only, but snippet should be enough: there's no default endpoint and we leave as it is

computeralex92: With https://github.com/dutchcoders/transfer.sh/pull/244/commits/cb17b4a1ccc5c521b540b5ba1912c86239558d66 the default endpoint is removed; only the region is set with a default value.

aspacca:
I totally forgot about default value from `Cmd`, thanks for resolving in a much more elegant way

PR #249, opened 2019-07-21
Add support for S3 path style URLS.

n8225:
Current support for Minio and possible other S3 providers is broken.  This PR adds the option to utilize path style URLs by setting `s3-path-style` to `true`, `s3-path-style` defaults to `false`.


aspacca:
@n8225 thanks I don't use S3 and I don't have visibility about all this stuff

was broken since the beginning or since we moved to aws sdk?
can you explain what's a path style URL?

n8225:
Just since the move to the aws sdk.

PR #250, opened 2019-07-21
Fix s3 100MB max upload issue.

n8225:
By setting `u.MaxUploadParts = concurrency` and `u.PartSize = (1 << 20) * 5`, there was a 100MB limit on S3 uploads.  This PR removes that limit leaving MaxUploadParts at the default of 10,000 and allowing the option to set the part size for S3 uploads with `s3-part-size`.

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -43,6 +43,7 @@ func getAwsSession(accessKey, secretKey, region, endpoint string, forcePathStyle
 		Endpoint:         aws.String(endpoint),
 		Credentials:      credentials.NewStaticCredentials(accessKey, secretKey, ""),
 		S3ForcePathStyle: aws.Bool(forcePathStyle),
+		//LogLevel:         aws.LogLevel(aws.LogDebug),
 	}))
 }
 
--- a/server/utils.go
+++ b/server/utils.go
@@ -43,7 +43,6 @@ func getAwsSession(accessKey, secretKey, region, endpoint string, forcePathStyle
 		Endpoint:         aws.String(endpoint),
 		Credentials:      credentials.NewStaticCredentials(accessKey, secretKey, ""),
 		S3ForcePathStyle: aws.Bool(forcePathStyle),
-		//LogLevel:         aws.LogLevel(aws.LogDebug),
 	}))
 }
 
--- a/README.md
+++ b/README.md
@@ -165,6 +165,7 @@ aws-secret-key | aws access key | | AWS_SECRET_KEY
 bucket | aws bucket | | BUCKET
 s3-region | region of the s3 bucket | eu-west-1 | S3_REGION
 s3-no-multipart | disables s3 multipart upload | false | |
+s3-part-size | Size of parts for S3 multipart upload. | 5(MB) | |
 s3-path-style | Forces path style URLs, required for Minio. | false | |
 basedir | path storage for local/gdrive provider| |
 gdrive-client-json-filepath | path to oauth client json config for gdrive provider| |
--- a/README.md
+++ b/README.md
@@ -163,9 +163,9 @@ provider | which storage provider to use | (s3, grdrive or local) |
 aws-access-key | aws access key | | AWS_ACCESS_KEY
 aws-secret-key | aws access key | | AWS_SECRET_KEY
 bucket | aws bucket | | BUCKET
+s3-endpoint | Custom S3 endpoint. | |
 s3-region | region of the s3 bucket | eu-west-1 | S3_REGION
 s3-no-multipart | disables s3 multipart upload | false | |
-s3-part-size | Size of parts for S3 multipart upload. | 5(MB) | |
 s3-path-style | Forces path style URLs, required for Minio. | false | |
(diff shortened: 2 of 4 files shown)


aspacca:
@n8225 the problem is not clear on my side.
probably I got the wrong implication of setting `u.MaxUploadParts`: I wanted to replicate the `s3-no-multipart` flag behaviour with the new use of the aws sdk.
maybe it doesn't make sense at all to have this flag anymore. we had a specific PR for this (see: https://github.com/dutchcoders/transfer.sh/pull/201/files)

do you know if it is still a valid problem? in this case I would rather drop any extra flags and let the upload in the aws sdk take care of everything.

also to fix the issue of limiting the max size amount I see only dropping setting `u.MaxUploadParts` should be enough. I wouldn't go in being able to tweak the different packages behaviour through flags if it's not really needed

n8225:
@aspacca It looks like we could remove setting both `u.MaxUploadParts` and `u.PartSize` then the uploader would adjust as needed based on the file size.

I am not sure if `s3-no-multipart` is still needed maybe @stefanbenten could chime in on this one.  I know current versions of Minio support multipart uploads and it looks like it's the default for AWS S3, but there is a long list of S3 providers.

stefanbenten:
@n8225 @aspacca s3-no-multipart is still needed for the communication to the S3-Gateway of the Storj Platform (https://github.com/storj/storj). Currently it does not support full multipart uploads, which means that until that is implemented, we are unable to remove the flag.

That said, i am working on a native driver to upload and download without having the S3-Gateway in the middle.

aspacca:
@stefanbenten have you tested if current mater works with the S3-Gateway of the Storj Platform?
do you have time to do in case? I'd would be appreciated

We removed the old library we were using, that explicitly called multi parts upload, with the the Uploader from the official AWS SDK. This could have break the support: depending on the results different solution has to be taken in this branch

thanks

aspacca:
btw @stefanbenten are you an official developer at storj? does this mean that storj is currently using the upstream from this repo for https://transfer.sh? (it doesn't seem so)

stefanbenten:
@aspacca Yes i am. I am not using the upstream version. The last version has been shipped by @nl5887  to me.

RE testing: I definitely can test the latest binaries, if you let me know where to find them.

aspacca:
@stefanbenten can you compile from latest master? otherwise what architecture do you need the binary for?

stefanbenten:
@aspacca Would have to look whats necessary to also get the frontend. Last time i tried i was missing that.
But besides that, i would need GOOS=linux and GOARCH=amd64 👍

aspacca:
@stefanbenten building from source will fetch the `transfer.sh-web` dependency for the frontend. it will be the one from master and not the one you have on storj hosted service

aspacca:
according to this: https://github.com/aws/aws-sdk-go/issues/2439, the current implementation should work, but for the bug related to setting `u.PartSize` as well

aspacca:
@stefanbenten https://github.com/dutchcoders/transfer.sh/releases/download/v1.1.1/tranfer-sh.v1.1.1-linux-amd64.tgz
not tested: I cross-compiled from osx

aspacca:
@stefanbenten any chance to test? :)

aspacca:
@n8225 since we have no feedback from @stefanbenten I'd go for only dropping setting `u.MaxUploadParts`. do you want to do in this PR? otherwise I'd close this and do

stefanbenten:
@aspacca I am sorry, i was very busy the last few days.
Trying to add this test to my ToDo today, will report back as soon as i have results.

stefanbenten:
@aspacca Finally had time to test it across the weekend. We still require the flag to be persistent or at least my tests failed when it was not set.
I am writing a native binding to the Storj Network in the upcoming weeks, and will remove that setting afterwards. Sounds like a plan?

aspacca:
@stefanbenten thank you. on my side the doubt was not about still needing or not the flag, but rather than switching for the previous library for s3 to the official go sdk when using the label it wouldn't work as expected anymore.
so, this mean we only need to get rid of setting `u.MaxUploadParts`
@n8225 thank you also for your patient waiting a proper feedback for the PR. if you won't have time to push a new commit with the needed change by the end of the week I will try to do: you will be anyway referenced in the release

server/utils.go
```diff
@@ -43,6 +43,7 @@ func getAwsSession(accessKey, secretKey, region, endpoint string, forcePathStyle
 		Endpoint:         aws.String(endpoint),
 		Credentials:      credentials.NewStaticCredentials(accessKey, secretKey, ""),
 		S3ForcePathStyle: aws.Bool(forcePathStyle),
+		//LogLevel:         aws.LogLevel(aws.LogDebug),
```
aspacca: please, remove commented line :)

aspacca:
all good @n8225 , thanks. once removed the commented line I will merge

server/utils.go
n8225: Commented line removed.

n8225:
Commented line removed.

PR #251, opened 2019-07-29
Fix fish alias for busybox systems

different55:
While GNU mktemp will handle any number of X's, busybox mktemp (like in Android/Termux and distros like Alpine) insists on exactly 6 of them for whatever reason.


aspacca:
thanks @different55
actually BSD mktemp behave differently:
```
     mktemp [-d] [-q] [-t prefix] [-u] template ...
     mktemp [-d] [-q] [-u] -t prefix
```

Issue #257, opened 2019-08-22
Couldn't upload file to docker


ichcc:
Docker run as recommended

```bash
$ docker run -p 8080:8080 dutchcoders/transfer.sh:latest --provider local --basedir /home/ichcc/tmp
[transfer.sh]2019/08/22 12:31:11 Transfer.sh server started.
using temp folder: /tmp/
using storage provider: local
[transfer.sh]2019/08/22 12:31:11 listening on port: :8080
[transfer.sh]2019/08/22 12:31:11 ---------------------------
2019/08/22 12:31:15 mkdir /home/ichcc/tmp/NMb6L: no such file or directory
[transfer.sh]2019/08/22 12:31:16 172.17.0.1:44804 - - [2019-08-22T12:31:16Z] "PUT /RH.txt HTTP/1.1" 500  "" "curl/7.52.1"
^C[transfer.sh]2019/08/22 12:32:11 Server stopped.
```
Basedir exist, full access

```bash
$ curl --upload-file  ./torrents/RH.txt http://127.0.0.1:8080/RH.txt --verbose -H "Max-Days: 1"
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> PUT /RH.txt HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.52.1
> Accept: */*
> Max-Days: 1
> Content-Length: 3340
> Expect: 100-continue
>
* Done waiting for 100-continue
* We are completely uploaded and fine
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< Server: Transfer.sh HTTP Server 1.0
< X-Content-Type-Options: nosniff
< X-Made-With: <3 by DutchCoders
< X-Served-By: Proudly served by DutchCoders
< Date: Thu, 22 Aug 2019 12:32:19 GMT
< Content-Length: 24
< Connection: close
<
Could not save metadata
* Curl_http_done: called premature == 0
* Closing connection 0
```
```bash
$ docker info
Client:
 Debug Mode: false

Server:
 Containers: 21
  Running: 1
  Paused: 0
  Stopped: 20
 Images: 10
 Server Version: 19.03.1
 Storage Driver: overlay2
  Backing Filesystem: extfs
  Supports d_type: true
  Native Overlay Diff: true
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
[...]

PR #260, opened 2019-08-23
Fixes "mkdir /basedir: no such file or directory"

dopessoa:
If the basedir does not exist (like the case described on ISSUE #257), the Mkdir call fails.

Just changed to MkdirAll which creates the directories recursively.


aspacca:
👍 thanks @dopessoa

PR #265, opened 2019-08-29
Integrate continuous fuzzing

### LINKED ISSUE #253: Suggestion: Continuous Fuzzing
Hi, I'm Yevgeny Pats Founder of [Fuzzit](https://fuzzit.dev) - Continuous fuzzing as a service platform.

We have a free plan for OSS and I would be happy to contribute a PR if that's interesting.
The PR will include the following
- [go-fuzz](https://github.com/dvyukov/go-fuzz) fuzzers (This is generic step not-connected to fuzzit)
- Continuous Fuzzing of master branch which will generate new corpus and look for new crashes
- Regression on every PR that will run the fuzzers through all the generated corpus and fixed crashes from previous step. This will prevent new or old bugs from crippling into master.

You can see our basic example [fuzzitdev/example-go](https://github.com/fuzzitdev/example-go) and you can see an example of "in the wild" integration [google/syzkaller](https://github.com/google/syzkaller).

Let me know if this is something worth working on.

Cheers,
Yevgeny

bookmoons:
Integrates with Fuzzit for automated bug discovery. Initially discussed in #253.

I've tried to hit all publicly exposed channels. This fuzzes the profile listener, HTTP listener, and HTTPS listener. I had to get the listeners accessible through `Server` so it's possible to shut them down each iteration. They're private fields to keep them internal.

A local fuzzing run hit a segfault on all 3 if you send just a `0`. I'm attaching the data here. The unsuffixed file contains the crashing input.

[crashers.zip](https://github.com/dutchcoders/transfer.sh/files/3554293/crashers.zip)

There's a `transfersh` org on Fuzzit all preconfigured and ready to go. If you [sign in](https://fuzzit.dev/) to set up an account I can ask them to add you to the org. There's a little setup to do:
- In Fuzzit settings grab an API Key.
- In repo settings in [Travis](https://travis-ci.org/) paste it to envvar `FUZZIT_API_KEY`.

Thank you for considering.

Closes #253.


aspacca:
@bookmoons thanks for the branch
anyway the panic is due to improper setup of the Server struct not because of the input: `Server.storage` is missing/`nil`
and in case you try to run the binary without passing a correct storage type with the related mandatory params the binary will panic as well
also it's not so clear to me the value of sending raw bytes to the listener: it will test `net/http`, `net/http/pprof` and `github.com/gorilla/mux` packages not transfer.sh one, since if you don't perform the http request fitting one of the the handlers of transfer.sh you will always hit `Server.notFoundHandler`

I guess the proper fuzz test should be executed on the different `Storage` implementations. for `LocalStorage` that would be possible without much effort. for `S3Storage` we can setup a minio instance, but for the `GDrive` and upcoming `Storj` there's only the option to use the real service as far as i know.

bookmoons:
Thanks for looking at it @aspacca. Will look into these changes.

aspacca:
@bookmoons I can try to build a simple test scenario for `Local` storage now that I understood how to build the test

bookmoons:
That would be very helpful.

aspacca:
@bookmoons signed in at fuzzit
I see my branch fails regression because it looks for previous Fuzz functions declared in your branch: what is the way to reset this?

bookmoons:
> @bookmoons signed in at fuzzit

Put in that request to get you into the org.

PR #266, opened 2019-08-29
Implement Native Storj Storage Adapter

stefanbenten:
With this PR transfer.sh has native bindings to the Storj network.
Previously it was using the S3 backend and proxying the data through a S3 Gateway ( https://github.com/storj/storj/cmd/gateway/ ).

@aspacca If needed, i can grant access to the Beta Network to try it out. I would appreciate a throughout test!


aspacca:
@stefanbenten thanks for the effort
> i can grant access to the Beta Network to try it out. I would appreciate a throughout test!
for sure!

server/storage.go
```diff
@@ -17,11 +17,14 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+	"github.com/zeebo/errs"
```
aspacca: I would rather prefer not to use any third party package for errors

server/storage.go
```diff
@@ -561,3 +564,113 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+
+func NewStorjStorage(endpoint, apiKey, bucket, encKey string, logger *log.Logger) (*StorjStorage, error) {
+	var instance StorjStorage
+	var err error
+
+	ctx := context.TODO()
+
+	instance.uplink, err = uplink.NewUplink(ctx, nil)
+	if err != nil {
+		return nil, uplinkFailure.Wrap(err)
```
aspacca: no real needs in my opinion for all these errors wrapping. I would rather return the given `err` as returned by the sdk everywhere

aspacca:
@stefanbenten but for the observation about `errors` it seems fine and neat.
I will try to run some tests once I'll have access to beta network

I have one question: there is a way to bring up a storj uplink locally for CI purpose?
I plan to add integration tests for every `Storage` type but I'm blocked by no local/development env for testing `GDrive`. will be it the same for `Storj`?

stefanbenten:
> @stefanbenten but for the observation about `errors` it seems fine and neat.
> I will try to run some tests once I'll have access to beta network
> 
> I have one question: there is a way to bring up a storj uplink locally for CI purpose?
> I plan to add integration tests for every `Storage` type but I'm blocked by no local/development env for testing `GDrive`. will be it the same for `Storj`?

The easiest way to test locally, within CI is our local testnetwork ( https://github.com/storj/storj/wiki/Test-network ).

Regarding the errors package, we like to introduce the package to make log parsing easier.
But if requested, i can remove it of course!

stefanbenten:
Any updates? @aspacca
Let me know if i can help with anything!

aspacca:
@stefanbenten sorry. a bit busy lately: I will try to test it in the weekend

aspacca:
> Regarding the errors package, we like to introduce the package to make log parsing easier.
> But if requested, i can remove it of course!

regarding this: it would be appreciated if you can remove and use a prefix in the error message for easing your log parsing. what do you think?

later on we can think about a PR to armonize error wrapping all over the code

aspacca:
@stefanbenten tried to build the local testnerwork today and got the following on go 1.12.7:
`# storj.io/storj/internal/dbutil/sqliteutil
internal/dbutil/sqliteutil/migrator.go:59:15: srcConn.Raw undefined (type *sql.Conn has no field or method Raw)
internal/dbutil/sqliteutil/migrator.go:65:18: destConn.Raw undefined (type *sql.Conn has no field or method Raw)
s`
fyi

aspacca:
@stefanbenten sorry. it seems that your branch doesn't work with local testnetwork or some prerequisites are missing.
when I try to connect to the local testnetwork (after figuring out that the `STORJ_API_KEY` is actually the macaroon serialization of `GATEWAY_0_ACCESS_KEY`) the process panics with the following:
`panic: storj failure: metainfo error: transport error: connection error: desc = "transport: authentication handshake failed: tls peer certificate verification error: not signed by any CA in the whitelist: CA cert"`

aspacca:
> after figuring out that the STORJ_API_KEY is actually the macaroon serialization of GATEWAY_0_ACCESS_KEY

this may be even wrong, but I don't find any documentation about this for the local testnetwork

aspacca:
@stefanbenten even with `--storj-skip-peer-ca` still the same error

aspacca:
@stefanbenten I confirm: still failing with the same error even with `--storj-skip-peer-ca`

aspacca:
@stefanbenten I fixed the CI issues on travis in current master, you can rebase from it in case

aspacca:
@stefanbenten any news on this?

stefanbenten:
@aspacca Just pushed the changes! It now has the flag to allow testing against the local testnetwork. I missed to push the actual config change into the uplink..

aspacca:
sorry @stefanbenten I'm not able to run your branch without getting an `Unauthorized API credentials` error
Is very uncleary to me the right credentials to use for `--storj-apikey` and `--storj-enckey`
the first seem to be a macaroon serialized api key, but is it the key `GATEWAY_0_ACCESS_KEY` from `storj-sim network` env? (and why not doing the macaroon serialization in the code and just pass the plain value?)
the second I supposed to be `GATEWAY_0_SECRET_KEY` therefore, since they are the only credentials available in the env command
but I tried multiple times and I always get the `Unauthorized API credentials` error

stefanbenten:
@aspacca No worries. Let me prepare a setup script that should get you up to speed.
The encryption key can be chosen as you like (thats the key that is used to encrypt the data before it gets uploaded into the network (or in your case the storj-sim local network).
The `storj-apikey` should be the `GATEWAY_0_ACCESS_KEY` yeah.

I'll add the script here in a bit.

aspacca:
@stefanbenten what I should set as `--storj-scope` to test locally?

stefanbenten:
Hey @aspacca ! We are still working on the implementation to make it more robust and possible offering a client side thin client, that allows direct download without the need to pass it through the webserver.
I also noticed that during a download i am getting a nil pointer back, which is occuring somewhere in the handler.

To get the scope you have multiple ways. The easiest way is to open the config of the gateway and copy the scope from there.

stefanbenten:
Good news @aspacca , i just got all errors cleaned up and now the PR is ready to go.
I sent you a Email with Invites to our production network to give you more options for testing. I also gave further assistance for the scope in it.

Looking forward to get this merged!

stefanbenten:
@aspacca Just checking in, did you get any chance to test this?

aspacca:
@stefanbenten tested on local test server, everything seems good.
please, if you can remove the error wrapping I'm ready to merge :)

stefanbenten:
@aspacca Just pushed the changes. We are currently working on seperating the library out of our main repo to cut down its dependency list. ETA on this is within the first week of January. Then it should be good to merge!

aspacca:
@stefanbenten 
> ETA on this is within the first week of January. Then it should be good to merge!
any update on this?

server/storage.go
```diff
@@ -561,3 +564,115 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+	ctx := context.TODO()
+
+	writer, err := s.project.UploadObject(ctx, s.bucket.Name, key, &uplink.UploadOptions{})
+	if err != nil {
+		return err
+	}
+
+	n, err := io.Copy(writer, reader)
+	if err != nil {
+		return err
```
egonelbre: call writer.Abort, since the reader might have errored and you don't want:
1. leave stuff lying around on the satellite
2. leave connections open

server/storage.go
```diff
@@ -561,3 +564,115 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+	}
+
+	err = writer.SetMetadata(ctx, &uplink.StandardMetadata{ContentType: contentType, ContentLength: n}, nil)
+	if err != nil {
+		return err
```
egonelbre: call abort here

server/storage.go
```diff
@@ -561,3 +564,115 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+	}
+	//Commit the object!
+	err = writer.Commit()
+	return err
+}
+
+func (s *StorjStorage) IsNotExist(err error) bool {
+	return storj.ErrObjectNotFound.Has(err)
```
egonelbre: The errors changed

server/storage.go
```diff
@@ -561,3 +564,115 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+	}
+
+	instance.project, err = uplink.OpenProject(ctx, parsedAccess)
+	if err != nil {
+		return nil, err
+	}
+
+	instance.bucket, err = instance.project.EnsureBucket(ctx, bucket)
+	if err != nil {
+		return nil, err
```
egonelbre: close project

stefanbenten: done

server/storage.go
stefanbenten: done

server/storage.go
stefanbenten: Done

README.md
```diff
@@ -151,7 +153,35 @@ If you specify the s3-region, you don't need to set the endpoint URL since the c
...
+Once the uplink is setup create the bucket using the following schema: 
+`uplink mb sj://<BUCKET>` where <BUCKET> is your desired name.
+
+Afterwards you can copy the SCOPE out of the configuration file of the uplink and then start the startup of the
+transfer.sh endpoint. For enhanced security its recommended to provide both the scope and the bucket name as ENV Variables.
+
+Example:
+```
+export STORJ_BUCKET=transfersh
+export STORJ_SCOPE=<SCOPE>
```
kaloyan-raev: This README has to be updated to say "access" instead of "scope".

cmd/cmd.go
```diff
@@ -362,6 +374,16 @@ func New() *Cmd {
 			} else {
 				options = append(options, server.UseStorage(storage))
 			}
+		case "storj":
+			if access := c.String("storj-access"); access == "" {
+				panic("storj-scope not set.")
```
kaloyan-raev: Should be "storj-access not set."

server/storage.go
```diff
@@ -561,3 +564,124 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+
+	ctx := context.TODO()
+
+	obj, err := s.project.StatObject(ctx, s.bucket.Name, key)
+	if err != nil {
+		return "", 0, err
+	}
+
+	customMeta := obj.Custom.Clone()
+	contentType = customMeta["contentType"]
```
kaloyan-raev: MinIO uses "content-type" (not "contentType") as the key in metadata. Should it be the same here?

server/storage.go
```diff
@@ -561,3 +564,124 @@ func saveGDriveToken(path string, token *oauth2.Token, logger *log.Logger) {
...
+	key := storj.JoinPaths(token, filename)
+
+	ctx := context.TODO()
+
+	obj, err := s.project.StatObject(ctx, s.bucket.Name, key)
+	if err != nil {
+		return "", 0, err
+	}
+
+	customMeta := obj.Custom.Clone()
```
kaloyan-raev: Cloning the metadata map seems to be unnecessary here.

go.mod
```diff
@@ -36,6 +36,6 @@ require (
 	google.golang.org/genproto v0.0.0-20191007162740-aa923e3a3354 // indirect
 	google.golang.org/grpc v1.24.0 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15
-	storj.io/common v0.0.0-20200214070817-cfd197b055d7
-	storj.io/uplink v0.0.0-20200221171743-26268cdd3552
+	storj.io/common v0.0.0-20200226144507-3fe9f7839df5
+	storj.io/uplink v1.0.0-rc.2
```
egonelbre: We are already on rc.3

aspacca:
@stefanbenten I've refactored the return signature of `Get` and `Head` methods in the `Storage` interface, can you please update in your branch?
Now there's no need to fetch the content type in those methods only relying on info stored in the metadata.
once you've done I will do the final CR, merge to master and bump to `v1.2.0`
thanks

server/storage.go
```diff
@@ -665,5 +663,5 @@ func (s *StorjStorage) Put(token string, filename string, reader io.Reader, cont
 }
 
 func (s *StorjStorage) IsNotExist(err error) bool {
-	return uplink.ErrObjectNotFound.Has(err)
+	return err == uplink.ErrObjectNotFound
```
egonelbre: You need to use `errors.Is(err, uplink.ErrObjectNotFound)`.

stefanbenten:
@aspacca Done!

aspacca:
@stefanbenten all good, but CI haven't run since a while (I havent' checked why)
after a green run I can merge

aspacca:
@stefanbenten 👍

stefanbenten:
Not sure why travis failed.

aspacca:
@stefanbenten I've removed fuzzit tests in master, please rebase and it should be fixed

PR #267, opened 2019-08-31
Fuzz Local Storage

francescayeye:
@bookmoons this is a skeleton of what I think should look like the fuzz test
please review, and if it's ok I will properly randomize token and filename and all the rest

alternatively instead of testing directly the storage we can build the http request to the listener that will hit the equivalent method in the storage that I tested now: what would be the best solution?

btw: I didn't run the code yet


bookmoons:
> alternatively instead of testing directly the storage we can build the http request to the listener that will hit the equivalent method in the storage that I tested now: what would be the best solution?

I like the way you're heading with getting right at the storage method. Probably it will make fuzzing faster and get through more code.

bookmoons:
I wonder if some of these conditions should panic so they show up as crashes. Are these things that should never happen?

eg for Head:

```go
	contentType, contentLength, err := storage.Head("token", "filename")
	if err != nil {
		return 0
	}

	if contentType != "application/octet-stream" {
                panic("incorrect content type")
	}

	if contentLength != uint64(len(fuzz)) {
		panic("incorrect content length")
	}
```

bookmoons:
I think it's in good shape.

server/server_fuzz.go
```diff
@@ -0,0 +1,87 @@
...
+		return 0
+	}
+
+	err = storage.Delete("token", "filename")
+	if err != nil {
+		return 0
+	}
+
+	_, _, err = storage.Head("token", "filename")
+	if err != nil {
```
bookmoons: Maybe this err check is not needed. I think it prevents the IsNotExist check from happening.

aspacca: indeed

aspacca:
> I wonder if some of these conditions should panic so they show up as incorrect behavior. Are these things that should never happen?

yes, they should never happen.
can you explain me the ratio between panicking or returning 0 in the fuzz test?

bookmoons:
They're using go-fuzz. It does this genetic fuzzing where it watches which parts of the program each input executes and uses that to decide what to mutate next and what to keep the same. So there's this coverage logic as the default and you can give signals to modify it:

* `0` return means there's no special meaning to this case so use the normal coverage logic.
* `1` return means increase priority, they want that for a successful case that gives no errors.
* `-1` means force removal from the corpus, this is for invalid input you don't even want to run.
* Panicking means we've found some incorrect behavior. Either the program itself can panic or you can do it explicitly if you've detected wrong behavior externally.

There's the relevant bit from the go-fuzz readme:

>Data is a random input generated by go-fuzz, note that in most cases it is invalid. The function must return 1 if the fuzzer should increase priority of the given input during subsequent fuzzing (for example, the input is lexically correct and was parsed successfully); -1 if the input must not be added to corpus even if gives new coverage; and 0 otherwise; other values are reserved for future use.

>To communicate application-level bugs Fuzz function should panic (os.Exit(1) will work too, but panic message contains more info).

bookmoons:
Whoops, that build failure is my fault. Forgot to give the new target a public corpus. I've updated. If you build again it should work.

aspacca:
I see this in the regression log:
```Connecting to storage.googleapis.com (storage.googleapis.com)|173.194.197.128|:443... connected.
891HTTP request sent, awaiting response... 404 Not Found
8922019-09-01 07:20:44 ERROR 404: Not Found.```

is it fine?

bookmoons:
Yeah that's OK. It's because it doesn't have a corpus yet. Once fuzzing builds a corpus it will have something to run regression tests on.

bookmoons:
I see the main fuzzing is failing, because there's no API key. If you'd like to get it set up you could email me. I can send you a key from the org.

waic9e at protonmail dot com

aspacca:
green

bookmoons:
I see the fuzz job in Fuzzit. Looks like it found a crash. I'm uploading the crashing input, single byte file. The error is incorrect content type.

[artifact.zip](https://github.com/dutchcoders/transfer.sh/files/3563247/artifact.zip)

aspacca:
this mean that build is green even if it found crashes? not direct feedback in the build?

bookmoons:
Yeah, it happens like this:

* Fuzz regression test runs all past crashes. It runs in CI and fails the build in any of them fail.
* Main fuzzing submits to Fuzzit for a longer run and finishes the build as soon as the job is in the queue.

Fuzzit sends you an email when it finds a crash. Once you're in the org you'll start getting them.

aspacca:
I fixed the unexpected content type crash: actually this is some inconsistency in the `Storage` interface where some of the implementations behave differently according to the type and the input is ignored and/or the returned one is evaluated on the getter
I will think about aligning all of them in a single behaviour

bookmoons:
It's looking good. What do you think? Is this one ready to merge?

aspacca:
just removing now unneeded listeners exposure

PR #274, opened 2019-11-13
Remove explicit OS and Arch builds

eksoverzero:
This could allow multi-arch supporting using `docker buildx`. At the very least, it enables support for ARM by building on an ARM device.

References: https://docs.docker.com/engine/reference/commandline/buildx/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ WORKDIR /go/src/github.com/dutchcoders/transfer.sh
 ENV GO111MODULE=on
 
 # build & install server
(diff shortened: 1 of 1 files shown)


aspacca:
@eksoverzero I see `buildx` is still experimental and not advertised to use in production
I would like to understand the implication of not setting `GOARCH` and `GOOS` then build and pushing the image on the docker registry

mckaygerhard:
@aspacca in any case there's a arch hard coded build here! and must be removed!

aspacca:
@mckaygerhard no, the docker file is here to push build to docker hub, since I honestly don't know if the hardcoded OS and ARCH can be removed for that I cannot just remove

if you want to build for a different OS and ARCH you have the base docker file that you can update

PR #275, opened 2019-11-14
add google drive instruction in readme

thanakijwanavit:
just adding instruction on how to use gdrive option


aspacca:
thanks @thanakijwanavit

PR #298, opened 2020-04-06
server: Replace string/parse trick for cloning with more efficient option.

calebcase:


aspacca:
thanks @calebcase , while you are at it do you mind checking for `nil` return?

calebcase:
> thanks @calebcase , while you are at it do you mind checking for `nil` return?

NP, nil check added

aspacca:
@calebcase now also `getURL` can return `nil` and it should be checked anywhere.
I wonder if the `nil` check is needed at all, can ever `http.Request.URL` be `nil`?

calebcase:
@aspacca I think this was a problem before. `r.URL` has paths where it could be `nil`. If it is, then the old code would have segfaulted at `r.URL.String()` (https://play.golang.org/p/LZ45LIv6bB6).

It may be better to establish an invariant earlier in the request processing to avoid having to have nil checks throughout the code base.

aspacca:
according to https://github.com/golang/go/blob/master/src/net/http/request.go#L1052 the request will never reach the handler if the url cannot be parsed, so we can assume `r.URL` will not be nil

aspacca:
@calebcase still working on this?

aspacca:
@calebcase if there's no update I'm going to close the PR and implement personally

calebcase:
Yup, please do!

PR #307, opened 2020-05-03
add proxy-port for port override when running behind a proxy

### LINKED ISSUE #306: Cannot use behind proxy running on a different port
Right now, the `--proxy-path` parameter allows you to run this program behind a web proxy which changes the path of a request, however it does not have a way to take into account the fact that a proxy may also be running on a different port.

For example, if I run this on port 1234 with a web proxy pointing to it at path /ex which is listening on port 80, uploading a file results in the response url: `http://localhost:1234/ex/<token>/<file>`

This is incorrect, as to go through the proxy, the port must be 80

An option needs to be added (something like `--proxy-port`) which allows you to specify the port number of a proxy fronting this service.

cheeseandcereal:
Closes #306 

Useful when running behind a proxy that does not allow modifying requests (can't set custom Host header), which could include firewall/routing rules between ports.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -158,9 +158,9 @@ func (s *Server) previewHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	relativeURL, _ := url.Parse(path.Join(s.proxyPath, token, filename))
-	resolvedURL := resolveURL(r, relativeURL)
+	resolvedURL := resolveURL(r, relativeURL, s.proxyPort)
 	relativeURLGet, _ := url.Parse(path.Join(s.proxyPath, getPathPart, token, filename))
-	resolvedURLGet := resolveURL(r, relativeURLGet)
+	resolvedURLGet := resolveURL(r, relativeURLGet, s.proxyPort)
 	var png []byte
 	png, err = qrcode.Encode(resolvedURL, qrcode.High, 150)
 	if err != nil {
@@ -170,8 +170,8 @@ func (s *Server) previewHandler(w http.ResponseWriter, r *http.Request) {
 
 	qrCode := base64.StdEncoding.EncodeToString(png)
 
-	hostname := getURL(r).Host
-	webAddress := resolveWebAddress(r, s.proxyPath)
+	hostname := getURL(r, s.proxyPort).Host
+	webAddress := resolveWebAddress(r, s.proxyPath, s.proxyPort)
 
 	data := struct {
 		ContentType   string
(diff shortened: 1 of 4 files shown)


server/handlers.go
```diff
@@ -555,16 +555,25 @@ func getURL(r *http.Request) *url.URL {
...
 	}
 
-	if u.Host != "" {
-	} else if host, port, err := net.SplitHostPort(r.Host); err != nil {
-		u.Host = r.Host
-	} else {
-		if port == "80" && u.Scheme == "http" {
-			u.Host = host
-		} else if port == "443" && u.Scheme == "https" {
+	if u.Host == "" {
```
aspacca, line 558: I should test it, but this will work only in case of tcp proxy/firewall where no `Host` is set. In case of an http proxy where we set the `Host` header the `--proxy-port` option would be ignored

aspacca: or in general if the client send an `Host` header

server/handlers.go
```diff
@@ -500,15 +500,15 @@ func (s *Server) putHandler(w http.ResponseWriter, r *http.Request) {
 	relativeURL, _ := url.Parse(path.Join(s.proxyPath, token, filename))
 	deleteURL, _ := url.Parse(path.Join(s.proxyPath, token, filename, metadata.DeletionToken))
 
-	w.Header().Set("X-Url-Delete", resolveURL(r, deleteURL))
+	w.Header().Set("X-Url-Delete", resolveURL(r, deleteURL, s.proxyPort))
 
-	fmt.Fprint(w, resolveURL(r, relativeURL))
+	fmt.Fprintln(w, resolveURL(r, relativeURL, s.proxyPort))
```
aspacca, line 505: please, keep `fmt.Fprint`
this is a long establihed behaviour of transfer.sh that returned url is not newline trailed

cheeseandcereal: Sure I can revert it. Interesting to note that it already is Fprintln for the POST handler, but not the PUT handler which seems a bit conflicting; with that said this PR obviously isn't trying to address that, so I can change it back

server/handlers.go
cheeseandcereal: In my testing, `u.Host` was _always_ empty regardless of how I sent the request to the server.

After further investigation, `r.URL.String()` was always either an empty string or simply "/" 100% of the time, regardless of what I was doing. This lead to the Host never being parsed into `u` by default regardless. The only place where the host seems to be defined is on `r.Host`, but NOT on `r.URL`

If you can find a case where `r.URL.String()` would _ever_ have host information, I would be curious, as I could never find a case where that was true. I don't think that the gorilla mux router includes host information on the URL of *http.Request that it passes into handlers; I believe it only provides host information on the *http.Request struct itself, but I could be wrong there.

The only reason I left in that if statement was because that was the behavior beforehand, although I currently don't believe that's ever actually used.

aspacca: @cheeseandcereal I'm going to do some test next week.
what kind of test did you try?

cheeseandcereal: I tried sending http requests to the server doing all of the following

- submitting requests with no Host header
- submitting requests with a normal Host header
- submitting requests with a Host header with a combined port (example.com:1234)
- doing all of the above on both the http and tls listeners

None of those cases resulted in host information on `r.URL`

PR #317, opened 2020-06-14
ci: deploy multi-architecture Docker images using buildx

### LINKED ISSUE #313: Multi-architecture Docker image
Hello,

I have been using the `transfer.sh` Docker image for quite some time now.

I would like to use it on arm64 but the image you  publish is only amd64.

I might be wrong about this but from what I saw, supporting other architectures should be pretty straightforward and easy using [`buildx`](https://docs.docker.com/buildx/working-with-buildx/) as you don't rely on other images than the `golang` one and that one is already multi-architecture.

Normally, you shouldn't even have to touch the source code or Dockerfile, just instruct Docker to build a multi-platform image and every thing else should be the same.

zuh0:
This PR adds a GitHub action to build and deploy a multi-architecture Docker image to Docker hub. In order for this to work, 2 secrets need to be added to the `dutchcoders/transfer.sh` GitHub repository.

The first one called DOCKER_USERNAME should be the Docker hub username for the "dutchcoders" account.

The second one called DOCKER_PASSWORD should be an access token generated on Docker hub like explained [here](https://docs.docker.com/docker-hub/access-tokens/).

The images will be built for the following architectures:

* amd64 (x86_64)
* i386 (x86)
* armv8 (aarch64/arm64)
* armv7 (arm32)

The images will be built (and sometimes pushed) in the following circumstances:

* on pull requests (just built)
* on tag pushes when the tag matches: vX.Y.Z (built and pushed as `dutchcoders/transfer.sh:vX.Y.Z`)
* on pushes to master (built and pushed as `dutchcoders/transfer.sh:edge` and `dutchcoders/transfer.sh:latest`)
* every night at midnight UTC to pull in updates in the dependencies and the `golang` image (built  and pushed as `dutchcoders/transfer.sh:nightly` and `dutchcoders/transfer.sh:latest`)

This PR resolves #313.

### DIFF
--- a/.github/workflows/build-docker-images.yml
+++ b/.github/workflows/build-docker-images.yml
@@ -1,0 +1,78 @@
+name: Deploy multi-architecture Docker images for transfer.sh with buildx
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # everyday at midnight UTC
+  pull_request:
+    branches: master
+  push:
+    branches: master
+    tags:
+      - v*
+
+jobs:
+  buildx:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Prepare
+        id: prepare
+        run: |
+          DOCKER_IMAGE=dutchcoders/transfer.sh
+          DOCKER_PLATFORMS=linux/amd64,linux/arm/v7,linux/arm64,linux/386
+          VERSION=edge
+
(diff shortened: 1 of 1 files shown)


aspacca:
@zuh0 thanks, I've tested your workflow on my branch removing the skip on pull requests so that we can test the built images
can you test dutchcoders/transfer.sh:edge? thanks

zuh0:
hello @aspacca , 

thanks for taking the time to handle this :+1: 
I just tested the amd64, armv7 and arm64 images and they work fine !

PR #330, opened 2020-10-12
Include version string in version command output

stillwondering:
The command `transfer.sh version` does not include the application's
version string in its output. Fix this.


aspacca (approved):

aspacca:
thanks @stillwondering

Issue #100, opened 2017-10-13
[Security Issue] Docker containers' processes run as root


TheAssassin:
(This is the proposal mentioned in https://github.com/dutchcoders/transfer.sh/issues/85#issuecomment-336584896.)

At the moment, the processes in the Docker containers run as root. This is a security problem, as Docker's isolation is not perfect and there's a lot of exploits that could make a root user break out of the container (it's way less of an issue for some random users which are ideally not existing on the host system).

Some links I picked up quite some while ago:

* http://blog.dscpl.com.au/2015/12/don-run-as-root-inside-of-docker.html
* https://stackoverflow.com/questions/41991905/docker-root-access-to-host-system

I don't have the time to fix and test this right now. PRs welcome, I guess.

Some drawbacks I experienced earlier which I don't want to withhold (might be useful for the person creating the fix):

 * When changing the user ID of the processes in the container, the existing volumes (whether it's a named one or a bind mount to a local directory) will require a manual upgrade (changing the owner)
 * Users trying to use a mount bind might experience issues (especially when that mount is already created but its owner is not set to the Docker process's user ID)

(If I should remember any other ones, I'll post them here.)

You can make it easier for anyone if you pick a very high user ID for the Docker container processes, anything above 10000 should suffice.

stek29:
since there are no complex entry points, it should be trivial to set uid/gid manually when starting the container
imo it should be up to end user to run the image properly, and images should not enforce uid/gid on their own

stek29:
I think that updating readme with note about setting uid/gid when starting the container should be enough

aspacca:
@stek29 if you are going to suggest how to update the readme I'm happy to do it (you can do as well opening a PR)

TheAssassin:
> imo it should be up to end user to run the image properly, and images should not enforce uid/gid on their own

That's total nonsense. A Docker image should not be insecure by default. It's considered a best practice to avoid running as root when it's totally unneeded. You can always override the uid/gid, but that's nothing you should have to do to make using a Docker image safe.

Docker environments aren't as isolated as, say, a VM. root can do a lot of harm, even from a container.

Implementing what I suggested is really trivial and increases safety a lot. I would've done it months ago. However, I've never seen a reply like, "ok, let's do it" or "would you be willing to do it".

https://engineering.bitnami.com/articles/why-non-root-containers-are-important-for-security.html
[...]

PR #345, opened 2021-02-16
Run app with non-root user

TheAssassin:
Running applications in Docker as root user opens various attack vectors, as Docker's "isolation" isn't meant to be the only safe guard against adversaries. It should be considered a "last resort". It's best practice to use a non-root user wherever possible.

This commit changes back to Alpine as a base, as it provides the adduser tool via busybox. It increases the image size by a few MiB (2-3, currently). One could optimize the image size by writing a custom `/etc/passwd`, but this seems quite unnecessary, especially given that Alpine is the base of so many other images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,8 +14,12 @@ ENV GO111MODULE=on
 # build & install server
 RUN go get -u ./... && CGO_ENABLED=0 go build -ldflags -a -tags netgo -ldflags '-w -extldflags "-static"' -o /go/bin/transfersh github.com/dutchcoders/transfer.sh
 
-FROM scratch AS final
+FROM alpine AS final
 LABEL maintainer="Andrea Spacca <andrea.spacca@gmail.com>"
+
+# Creating a non-root user to run the software with
+RUN adduser --system app
+USER app
 
 COPY --from=build  /go/bin/transfersh /go/bin/transfersh
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt


aspacca:
@TheAssassin having an alpine based image running a static binary as non-root user being less open to various attack vector srather than having a scratch image run as non-root user on the host system seems rather counter-intuitive.
can you explain why it should be safer?

TheAssassin:
The key term is "privilege escalation". Running as `root` is far more dangerous than having some tools in the userland.

All attacks that manage to escape out of a Docker container have one requirement: `root`. If you have `root`, you can do a _lot_ more stuff than you can do as a normal, unprivileged user: run syscalls to `mount` things, for instance, overwrite stuff on block devices, ...
Of course, modern systems have mitigation techniques in place. Docker containers are usually run with only a minimal set of cgroups capabilities (e.g., `SYS_ADMIN` is usually not permitted for good reasons), and there are sandboxing tools like AppArmor that also restrict the processes. All these, however, are prone to errors: configuration mistakes, bugs in the code, there's a lot that could go wrong. In the end, they're just meant as safety guards, and the more the better. Relying only on those is, however, not a good idea.

The worst I can think of with adding some userland tools is that, if a user manages to escape the Go process, e.g., through some shellcode (it doesn't use any form of libc, which already closes some attack vectors), or some bug in the Go compiler, they're left with just a busybox instance, a bunch of libraries and nothing else, as an unprivileged user. No matter what, they can't run privileged operations like mounting devices.

Millions of containers use Alpine as the base, and I am not aware of a big wave of attacks on those. Of course, that doesn't have to mean anything. But I also can't think of an attack vector that opens because some tools are laying around on the filesystem. I consider `root` access significantly more dangerous than busybox. And I'm convinced the security community agrees there.

TheAssassin:
By the way, I already offered a way to implement #100 without using Alpine, by just writing the `passwd` file ourselves. It's possible, but I don't see any real gain over just using Alpine as the base and using its tools, as explained in the commit message. If you feel better, though, we can also do that.

aspacca:
> By the way, I already offered a way to implement #100 without using Alpine, by just writing the `passwd` file ourselves

this could solve an issue with https://github.com/dutchcoders/transfer.sh/pull/345/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R21

unless the uid and gid of the newly created user are deterministic an user running the docker image and wanting to use a docker volume on the host for local storage cannot know what uid/gid give ownership to for the directory on the host.

writing the `passwd` file should solve the issue.

correct me if I'm wrong

TheAssassin:
> unless the uid and gid of the newly created user are deterministic

Good catch and good question. In my experience in the past years, the first system user to be created is always UID 100. The group assigned is `nogroup`. We can hardcode those values, if you want to, but I don't think it'll be necessary. Writing the file directly won't be any easier than adding `--uid`. That should not be the criterion you use to decide that question.

aspacca:
>we need to ensure that the files of people who upgrade the Docker container will be accessible

we cannot ensure this, the only thing we can do is to add some documentation/breaking update notes on the readme and in the release page. local storage set explicitely 700 for dir and 600 files permissions, so it will break for sure.

as for scratch vs alpine: 2-3MB more is 30-40% bigger image. I would go for scratch, there's also the consideration to reduce even further the attack surface

please let me know if it's fine for you

mckaygerhard:
i guess security is not considering here @aspacca  .. from scratch is just upstream withour any attention and sync from others, each piece of the base are independient of the other, using a base like alpine of whatever is far better cos is specialiced in that objetive.. is pretty clear and makes sense for any 4 finger of head!

using a better "from" will solve those stupid "breaks" .. @TheAssassin  has a good point here .. and the future will give us the reason..

aspacca:
@mckaygerhard either be based on alpine or scratch, running the transfer.sh binary inside the docker as an user instead of root will break the current installation with local storage

TheAssassin:
> 30-40% bigger image

First, this isn't quite true. The image is composed of multiple layers, one being the Alpine base layer. As mentioned in one of my previous comments, this base layer is in very widespread use, so the chance is high that it'll be used by any other image as well, and thus won't occupy additional space.
Also, who won't have the 2-3 MiB on their servers? This is a storage system, and I'm sure many many people use the `local` provider. Even just a few pictures are a lot larger than the base image. I don't think this is going to be an issue _at all_. It smells like premature optimization.

> @mckaygerhard either be based on alpine or scratch, running the transfer.sh binary inside the docker as an user instead of root will break the current installation with local storage

True, and that's an issue. I was thinking about adding a startup script that ensures the correct permissions, then launches the daemon with the right permissions...

mckaygerhard:
err pardon me, but as @TheAssassin  saids, is not breaking.. cos we can made pre-scripts to ensure correct permissions.. in fact linux from scratch has more "tune up" non standar scripts rather any other.. so.. i guess you are wrong XD

aspacca:
> It smells like premature optimization.

it would be premature optimization if we are going to build now the docker image for the first time.
we already have scratch based image, and moving to alpine is going to increase the size. assuming that alpine layer is already present in the image cache on the machine that will pull the image is just an assumption, and I won't go through that.

> I was thinking about adding a startup script that ensures the correct permissions, then launches the daemon with the right permissions...

this script cannot run that as root, otherwise it won't be able to ensures the correct permissions, so in the end we are letting back from the window what we are trying to get out of the door.
also I don't think that a docker image should take care of such task. if this is the option I would rather keep the dockerfile as it is and providing guidelines on how to run the container as non-root user on the host machine.

TheAssassin:
> so in the end we are letting back from the window what we are trying to get out of the door.

That's not the case. The actual process is still run as non-root. The shell script doesn't open ports or anything. In fact, you don't even have to keep the shell session open, you can just `exec` the Go stuff once you're done dropping privileges.

> also I don't think that a docker image should take care of such task

Of course it should. Easy and flawless upgrading is a core task of good Docker images. Database upgrades are also done by the init scripts in good containers. Automate what can be automated safely, reducing the risk that the user may screw up, and saving their own time.

> if this is the option I would rather keep the dockerfile as it is and providing guidelines on how to run the container as non-root user on the host machine.

Do you realize that this requires more than just adding `--uid` to the run command? A named volume's ownership is, upon the first mount, impacted by the `USER` command in the Dockerfile. You will have to, like, `sudo chown` the actual volume path on the host machine. Plus, it's a nonstandard setup.

I see a lot of reluctance on your end with regards to this pull request. Please decide if you want to improve the overall security by not running the process as `root` or not. Only then it makes sense to talk about issues like upgrading or maintaining a static user ID for the system user. All the facts have been presented.

aspacca:
> I see a lot of reluctance on your end with regards to this pull request. Please decide if you want to improve the overall security by not running the process as root or not

I have no reluctance, I just set the conditions for how it should be done: scratch based image, no permissions change script.

lkubb:
More an FYI since I'm mostly a Github lurker: I needed this functionality as well and patched it in with the scratch based image. The UID/GID are set at build time though. I intended it to be modifiable at the container creation stage, but that needs some logic before launching transfersh, which is hard for me to do without a shell (script) and some additional binaries (i.e. chown, su_exec/gosu/sue).

https://github.com/lkubb/transfer.sh/commit/d9fefd54919e997f0c2fbe8dd7f45c8ba06b1330

lkubb:
@aspacca Thank you, glad to hear. :)

In the meantime, I tried to build it on a different ("virgin") system, first without providing --build-arg. This surfaced another issue I was actually expecting (no files matching the glob `/tmp/useradd/*`), but that didn't happen before somehow. This is fixed now in https://github.com/lkubb/transfer.sh/commit/54e125f40e340afd75e4ffb81f2c23f67dc6728e (adding a placeholder file)

`COPY --from=build --chown=${RUNAS}` runs just fine without providing a value in the argument, same as `USER ${RUNAS}`. Both default to root, at least on the four systems I tried it with. It feels a little bit hacky, but works well.

My `docker-compose.yml`:
```yaml
---
version: '3.8'

services:
  transfer:
    build:
        args: {}
#            RUNAS: transfersh
#            PUID: 5001
#            PGID: 5001
        context: ./transfer.sh
    hostname: transfer
    environment:
        - PROVIDER=local
        - BASEDIR=/tmp
    ports:
      - "1234:8080"
    expose:
      - "8080"
```
first build with the above `docker-compose.yml`, second build with commented out args:
```console
$ git clone https://github.com/lkubb/transfer.sh
$ docker-compose up -d
Creating network "test_default" with the default driver
Building transfer
Step 1/18 : ARG GO_VERSION=1.16
Step 2/18 : FROM golang:${GO_VERSION}-alpine as build
1.16-alpine: Pulling from library/golang
a0d0a0d46f8b: Pull complete
31adcdaf11c8: Pull complete
b8b176561691: Pull complete
23d1c9dd6ab7: Pull complete
36803fd6ed32: Pull complete
Digest: sha256:78181bcf43be59a818e23095f21b3818f456895c3f1f2daaabdfd1af75cedd1f
Status: Downloaded newer image for golang:1.16-alpine
 ---> 1b35785aa3c4
Step 3/18 : RUN apk add git musl-dev
[...]
Step 14/18 : COPY --from=build --chown=${RUNAS}  /go/bin/transfersh /go/bin/transfersh
 ---> 52dcabc2af45
Step 15/18 : COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 ---> ea5dc64d26ba
Step 16/18 : USER ${RUNAS}
 ---> Running in 4b2d8756aad6
Removing intermediate container 4b2d8756aad6
 ---> d53b24877c39
Step 17/18 : ENTRYPOINT ["/go/bin/transfersh", "--listener", ":8080"]
 ---> Running in 104bb487a1f4
Removing intermediate container 104bb487a1f4
 ---> 760ee3ad7929
Step 18/18 : EXPOSE 8080
 ---> Running in b200625a2964
Removing intermediate container b200625a2964
 ---> 7dd773bdd234

Successfully built 7dd773bdd234
Successfully tagged test_transfer:latest
WARNING: Image for service transfer was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Creating test_transfer_1 ... done
$ docker-compose top
test_transfer_1
UID     PID    PPID    C   STIME   TTY     TIME                     CMD
---------------------------------------------------------------------------------------
root   20721   20701   0   20:36   ?     00:00:00   /go/bin/transfersh --listener :8080
$ vim docker-compose.yml
$ docker-compose up -d --build
Building transfer
Step 1/18 : ARG GO_VERSION=1.16
Step 2/18 : FROM golang:${GO_VERSION}-alpine as build
 ---> 1b35785aa3c4
Step 3/18 : RUN apk add git musl-dev
 ---> Using cache
 ---> 3645c13f2ffa
[...]
Step 14/18 : COPY --from=build --chown=${RUNAS}  /go/bin/transfersh /go/bin/transfersh
 ---> 3ea81e980658
Step 15/18 : COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 ---> b8d8ff3c9678
Step 16/18 : USER ${RUNAS}
 ---> Running in 3b9a99503cea
Removing intermediate container 3b9a99503cea
 ---> 6a13ed7763e8
Step 17/18 : ENTRYPOINT ["/go/bin/transfersh", "--listener", ":8080"]
 ---> Running in f1f29b24b49d
Removing intermediate container f1f29b24b49d
 ---> 2d6ae386dd65
Step 18/18 : EXPOSE 8080
 ---> Running in c153f0868bef
Removing intermediate container c153f0868bef
 ---> fb4ce9bb8f00

Successfully built fb4ce9bb8f00
Successfully tagged test_transfer:latest
Recreating test_transfer_1 ... done
$ docker-compose top
test_transfer_1
UID     PID    PPID    C   STIME   TTY     TIME                     CMD
---------------------------------------------------------------------------------------
5001   21197   21174   2   20:37   ?     00:00:00   /go/bin/transfersh --listener :8080
```

aspacca:
actual fix in https://github.com/dutchcoders/transfer.sh/pull/418

PR #346, opened 2021-02-16
README.md: Update Disclaimer

stefanbenten:
As I am now operator of the public instance living at https://transfer.sh, the disclaimer section is outdated and can be removed.


aspacca:
@stefanbenten you can add your reference for inquires related to the hosted service if you want

stefanbenten:
I'll draft up a new paragraph in a separate PR. Need to think about how to frame it 😄

PR #347, opened 2021-02-19
go get -u, go build ldflags, docker tag, inline handler

aspacca:


server/handlers.go
```diff
@@ -1009,6 +1009,15 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("X-Remaining-Downloads", remainingDownloads)
 	w.Header().Set("X-Remaining-Days", remainingDays)
 
+
+	if strings.Contains(contentType, "html") {
+		reader = ioutil.NopCloser(
+			bytes.NewReader(
```
stek29, line 1015: sorry, I've only now looked deeper into the code

what's the point of making a bytes.Reader out of SanitizeReader's .Bytes?
SanitizeReader returns a bytes.Buffer which conforms to io.Reader, so making a new reader from bytes seems weird to me

why not `reader = ioutil.NopCloser(bluemonday.UGCPolicy().SanitizeReader(reader))` ?

aspacca: you are right, I might have be mislead by the IDE :)
I see another problem: https://github.com/dutchcoders/transfer.sh/blob/master/server/handlers.go#L993

we call `Close()` on the original `io.ReadCloser`, but in the case we use the sanitizer we will call it on the `ioutil.NopCloser`, never closing the original one

what do you think about:
```
		sanitizedReader := bluemonday.UGCPolicy().SanitizeReader(reader)
		reader.Close()
		reader = ioutil.NopCloser(sanitizedReader)
```

stek29: > we call Close() on the original io.ReadCloser, but in the case we use the sanitizer we will call it on the ioutil.NopCloser, never closing the original one

that's what I thought too originally, but turns out it's not how go works
variables used in defer statement are resolved on defer statement "execution"

for example, following code prints X.Test and not Y.Test:
```go
package main

import "fmt"

type T interface {
	Test()
}

type X struct{}

func (X) Test() { fmt.Println("X.Test") }

type Y struct{}

func (Y) Test() { fmt.Println("Y.Test") }

func main() {
	var t T
	t = X{}
	defer t.Test()
	t = Y{}
}
```

see https://play.golang.org/p/7IIG3kbOcW6

or even more detailed example:
```go
func main() {
	var t T
	t = X{}

	defer t.Test()
	defer func() {
		t.Test()
	}()
	defer func(xt T) {
		xt.Test()
	}(t)

	t = Y{}
}
```

prints X, Y, X: https://play.golang.org/p/8hgd6wir391

also, you can't close original reader until sanitizereader is closed, since sanitizereader doesn't read whole file into memory

aspacca: so I will just unwrap the `bytes.Reader`

PR #362, opened 2021-04-19
Only rewrite metadata file when the download counter changed

JustAnotherArchivist:
Previously, the metadata file would be rewritten even if the download counter stayed the same (i.e. `increaseDownload = false`, previews and HEAD requests). Because the metadata doesn't change in that case, this would simply rewrite the exact same contents needlessly, which may also incur extra costs depending on the storage backend.


aspacca:
@JustAnotherArchivist thanks for the contribution and sorry for the late feedback :)

PR #368, opened 2021-05-01
GitHub workflow for releases

### LINKED ISSUE #367: Moving from travis-ci to github action - HELP WANTED :)
Travis-ci introduced a new billing policy (https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).
While CI and release artifcats seems to not be affected at the moment, the build badge is broken
([Build Status](https://travis-ci.com/dutchcoders/transfer.sh))
We should move from travis-ci to github action, adapting the current workflow: https://github.com/dutchcoders/transfer.sh/blob/master/.travis.yml

Any help would be appreciated

cc @stefanbenten

marzocchi:
You can see it in _action_ in my fork: a [run](https://github.com/marzocchi/transfer.sh/actions/runs/802287855) and a [release](https://github.com/marzocchi/transfer.sh/releases/tag/v1.2.3.4.5.6)

resolves #367


.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+            suffix: windows-amd64.exe
+          - GOOS: linux
+            GOARCH: amd64
+            suffix: linux-amd64
+          - GOOS: linux
+            GOARCH: arm
+            suffix: linux-armv7
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
marzocchi, line 28:
```suggestion
      - uses: actions/setup-go@v2
        with:
          go-version: '1.15'      
```

@aspacca you want a specific version?

aspacca: latest would be fine. it should have been upgraded to `1.16` on travis as well

.github/workflows/test.yml
```diff
@@ -0,0 +1,12 @@
+name: Release
+on: [push]
+jobs:
```
aspacca, line 3: missing go versions matrix

marzocchi: there might be some downside when using the job as status check for PRs, in that you'll have to select them all (and change the settings everytime you change the matrix)

Screenshot 2021-05-01 at 16 15 41

.github/workflows/release.yml
```diff
@@ -0,0 +1,85 @@
...
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        go_version:
+        - 1.13.x
+        - 1.14.x
+        - 1.15.x
+        - 1.16.x
```
aspacca, line 18: do we have access to `tip`?

aspacca (approved):

.github/workflows/test.yml
aspacca: I'll think about it

marzocchi:
@aspacca, to wrap it up: 

- i merged tests and build in a single file, with conditionals; you can [see here how it looks when pushing branches](https://github.com/marzocchi/transfer.sh/actions/runs/802395144) and [here when pushing tags](https://github.com/marzocchi/transfer.sh/actions/runs/802398344)

- you might want to suggest a better name that `Release` for the whole workflow 
Screenshot 2021-05-01 at 16 54 02

- regarding the matrix, you [should check if this is ok](https://github.com/dutchcoders/transfer.sh/pull/368#discussion_r624516694)

also [it looks like `action/setup-go` is a bit flaky when trying to use the latest version](https://github.com/actions/setup-go/issues/73), setting at least the minor version seems safer

.github/workflows/release.yml
marzocchi: does not look like, see my comment below; also for example I saw that earlier release builds, before I specified `1.16.x` where done with `1.15`

aspacca:
>i merged tests and build in a single file, with conditionals

indeed I would prefer different files, since we already have one for docker image

aspacca:
>also it looks like action/setup-go is a bit flaky when trying to use the latest version, setting at least the minor version seems safer

it seems that's a problem when wanting to set a minimum version defaulting to latest, that doesn't pick up really. if you pin a specific version it should be fine (but it needs maintance)

marzocchi:
> if you pin a specific version it should be fine (but it needs maintenance)

yup that's what I meant, and there's no keyword like `tip`

aspacca:
>yup that's what I meant, and there's no keyword like tip

got it now

marzocchi:
> indeed I would prefer different files, since we already have one for docker image

splitted!

aspacca:
can you change the badge in README.md as well?
thanks!

aspacca:
and delete `.travis.yaml`

marzocchi:
> can you change the badge in README.md as well?

done, but you won't see it until merge as the workflow in the main branch won't exist until then, but here's a preview from my fork.

---


# transfer.sh [Gitter](https://gitter.im/dutchcoders/transfer.sh?utm_source=badge&utm_medium=badge&utm_campaign=&utm_campaign=pr-badge&utm_content=badge) [Go Report Card](https://goreportcard.com/report/github.com/dutchcoders/transfer.sh) [Docker pulls](https://hub.docker.com/r/dutchcoders/transfer.sh/) [Build Status](https://github.com/dutchcoders/transfer.sh/actions/workflows/test.yml?query=branch%3Amaster)

---

how much are you annoyed by the difference in case on the black side of the badge? :D

aspacca:
>how much are you annoyed by the difference in case on the black side of the badge? :D

I will reject the PR, that's how much I'm annoyed ;P

marzocchi:
> >how much are you annoyed by the difference in case on the black side of the badge? :D
> 
> 
> 
> I will reject the PR, that's how much I'm annoyed ;P

Ah then I need to change the whole workflow name. But then it's inconsistent with the others, so I have to change them too. Including the existing one.

Maybe something breaks...!

PR #374, opened 2021-05-21
Bump Storj Dependencies

stefanbenten:
It has been a while and quite a few optimizations have been made to the libraries including striping out unnecessary dependencies.


aspacca (approved):

aspacca:
@stefanbenten it seems there's some problem with the test github actions

aspacca:
for the moment let's go with duplicated tests

PR #386, opened 2021-07-08
Multiple fixes

### LINKED ISSUE #353: Transfer.sh Deletion via browser returns 405 error, reconsider implementation for Delete via Get request?
Hi @stefanbenten ! I read through implementation discussion here: https://github.com/dutchcoders/transfer.sh/issues/44#issuecomment-399234697. 
So I understand it is not an issue with the code. It is the the website transfer.sh that should provide a delete form to process the delete GET-request, correct? So my wish would be that such a page would be set up for transfer.sh so that deletes can be successfully performed..

I'd also appreciate a reconsideration of that discussion. In my view a direct delete by a GET request (without DELETE header) should be made possible without the need for an additional form or button. To prevent accidental or malicious deletes via crawler the delete link should be constructed more randomly with longer passwords and should be different from the download links and not derived from them. I believe atm they are too similar and can be easily computed if the download link is known.

_Originally posted by @beriner in https://github.com/dutchcoders/transfer.sh/issues/351#issuecomment-783746108_

### LINKED ISSUE #377: Include handler errors in the log file
I realised today that many messages only get logged to stderr, not to the log file. In particular, this includes all errors in the handlers, e.g. errors on reading from the storage (i.e. the `log.Printf` calls in `handlers.go`). I'd like to suggest having those both on stderr and in the log file (if one is specified).

### LINKED ISSUE #380: Race condition crash in `CheckMetadata` locking: fatal error: concurrent map read and map write
Our transfer.sh server crashed today with the following traceback (only the relevant lines, paths redacted):

```
fatal error: concurrent map read and map write

goroutine 130672 [running]:
runtime.throw(0x104a276, 0x21)
        .../go1.15.11/src/runtime/panic.go:1116 +0x72 fp=0xc0007150a0 sp=0xc000715070 pc=0x4391b2
runtime.mapaccess2_faststr(0xed3c20, 0xc000449d40, 0xc000b148c0, 0x1d, 0x1d, 0xc000a8602a)
        .../go1.15.11/src/runtime/map_faststr.go:116 +0x4a5 fp=0xc000715110 sp=0xc0007150a0 pc=0x415fc5
github.com/dutchcoders/transfer.sh/server.(*Server).Lock(0xc000001680, 0xc000f2cab5, 0x5, 0xc000f2cabb, 0x17, 0x311d, 0x7231)
        .../transfer.sh/server/handlers.go:631 +0xc5 fp=0xc000715188 sp=0xc000715110 pc=0xda2665
github.com/dutchcoders/transfer.sh/server.(*Server).CheckMetadata(0xc000001680, 0xc000f2cab5, 0x5, 0xc000f2cabb, 0x17, 0xc000f2cb01, 0x0, 0x0, 0x0, 0x0, ...)
        .../transfer.sh/server/handlers.go:648 +0xd6 fp=0xc000715308 sp=0xc000715188 pc=0xda2936
github.com/dutchcoders/transfer.sh/server.(*Server).getHandler(0xc000001680, 0x11b5ca0, 0xc0009be720, 0xc000b0cb00)
        .../transfer.sh/server/handlers.go:991 +0x19c fp=0xc0007156d0 sp=0xc000715308 pc=0xda7efc
[...]

goroutine 130540 [runnable]:
github.com/dutchcoders/transfer.sh/server.(*Server).Lock(0xc000001680, 0xc000b263c5, 0x5, 0xc000b263cb, 0x17, 0x4edd97, 0xc000a96410)
        .../transfer.sh/server/handlers.go:632 +0x187
github.com/dutchcoders/transfer.sh/server.(*Server).CheckMetadata(0xc000001680, 0xc000b263c5, 0x5, 0xc000b263cb, 0x17, 0xc000b26401, 0x0, 0x0, 0x0, 0x0, ...)
        .../transfer.sh/server/handlers.go:648 +0xd6
github.com/dutchcoders/transfer.sh/server.(*Server).getHandler(0xc000001680, 0x11b5ca0, 0xc0001ca2d0, 0xc000580900)
        .../transfer.sh/server/handlers.go:991 +0x19c
[...]
```

The line numbers differ from the code in this repository as I'm running a modified code, but I'm all but certain that these modifications are unrelated to the crash. From the bottom, the corresponding lines in the current master code are:

https://github.com/dutchcoders/transfer.sh/blob/1594f95abf11fb9bb8e6e4581f6cc7e202370c1b/server/handlers.go#L986

https://github.com/dutchcoders/transfer.sh/blob/1594f95abf11fb9bb8e6e4581f6cc7e202370c1b/server/handlers.go#L660

https://github.com/dutchcoders/transfer.sh/blob/1594f95abf11fb9bb8e6e4581f6cc7e202370c1b/server/handlers.go#L643-L644

Note that the traceback is the same except for the one line difference in `Lock`, with one goroutine trying to create a mutex under `s.locks[key]` (line 632 in my code, 644 in master) and the other trying to access it (line 631 in my code, 643 in master).

This happened on a `GET` shortly after the upload of a new file. The upload was successful, and the URL was automatically shared and then accessed immediately around a dozen times in parallel and very quick succession (5 accesses within 100 ms, then another 5 in the following 200 ms plus a few more after, according to the Caddy logs). However, these kind of requests happen all the time on our instance and has never led to crashes before. transfer crashed before returning a response for any of the requests as far as I can tell.

I tried to reproduce this by running parallel requests against transfer (with a tool built in Python around aiohttp) but did not manage to trigger it that way. Perhaps the requests weren't quite quick enough. I haven't had time to set up and try other methods.

My interpretation is that this is a race condition in the locking code: two requests were trying to work with the metadata lock at the same time, one trying to create the mutex under `s.locks[key]`, the other trying to access that key, but because `s.locks` is a simple map, such simultaneous actions aren't possible.

francescayeye:
- fixes #380 
- always use logger - fixes #377 
- deletion in UI - fixes #353 
- more simple token generation
- version on build time
- configured purge time on ui


marzocchi (approved):

stefanbenten (approved):

stefanbenten (approved):

stefanbenten (dismissed):

stefanbenten:
Good to merge? @aspacca 
Likely worth a new release and shortly followed deploy 👍

aspacca:
@stefanbenten packed here also deletion in UI :)

yes, I will release `1.3.0` once merged

aspacca:
fixes #353

stefanbenten:
I'll test the version locally now 👍

stefanbenten:
@aspacca If you upload multiple files, the UI/UX looks a little odd.
image

Should we add a little more spacing or a divider to make it more clear that the token is for the previous URL, not the following?

JustAnotherArchivist:
I'm not a fan of such broad PRs, so I won't test this in full. But I have deployed the patch for #380 on our instance following another crash today. As mentioned in that issue, it's hard to properly test that this fixes the crash, but at the very least, it doesn't seem to break anything. :-)

aspacca:
@stefanbenten 
>Should we add a little more spacing or a divider to make it more clear that the token is for the previous URL, not the following?
>Additionally naming it Delete URL/Token might help?

both done

stefanbenten (approved):

PR #387, opened 2021-07-08
Added support for markdown-autodocs

dineshsonachalam:
## Problem with updating Code block or Markdown tables in your README.md 

To make your repo more appealing and, useful you need to provide Table or example code snippets in your README.md. Manually copy and pasting each code snippet or adding a table row in their respective places in your README would be inefficient and time-consuming.

## Solution to this problem:

This problem can be solved using markdown-autodocs GitHub Action that is created by me that automatically generates & updates markdown content (like your README.md) from external or remote files. You need to add markers in your README.md that will tell markdown-autodocs where to insert the code snippet or transform JSON to HTML table.

Markdown Autodocs GitHub action: https://github.com/dineshsonachalam/markdown-autodocs

To test or experiment markdown-autodocs without Github action: https://github.com/dineshsonachalam/markdown-autodocs#local-usage-without-github-action

Please let me know if any changes are required and will be happy to help.
cc: @nl5887

### DIFF
--- a/.github/workflows/markdown-autodocs.yml
+++ b/.github/workflows/markdown-autodocs.yml
@@ -1,0 +1,14 @@
+name: markdown-autodocs
+
+on:
+  push:
+    branches:    
+      - '**'        # matches every branch
+
+jobs:        
+  auto-update-readme:
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v2
+        - name: Markdown autodocs
+          uses: dineshsonachalam/markdown-autodocs@v1.0.2
--- a/README.md
+++ b/README.md
@@ -11,51 +11,42 @@ The service at https://transfersh.com is of unknown origin and reported as cloud
 ## Usage
 
 ### Upload:
-```bash
-$ curl --upload-file ./hello.txt https://transfer.sh/hello.txt
-```
+<!-- MARKDOWN-AUTO-DOCS:START (CODE:src=./examples/readme/upload.sh) -->
+<!-- MARKDOWN-AUTO-DOCS:END -->
 
 ### Encrypt & upload:
-```bash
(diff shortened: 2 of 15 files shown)


aspacca:
hi @dineshsonachalam this could be useful for maintaining `README.md` but indeed I see all the snippets are marked as `sh`, while they are `bash` specific. There should be `zsh` and `power shell` snippets as well if I remember correctly, but I don't see them migrated to the action.
Are they not supported?

dineshsonachalam:
Hi @aspacca

markdown-autodocs GitHub action is not limited to any file extensions (.py, .js, .go etc). It takes the extension of the file and add's it at the top of the code block to support syntax highlighting in the markdown file (like README.md).

Github supports syntax highlighting for the below languages.
Reference: https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

From the supported languages, we can also see .bash, .zsh, .ps1 (PowerShell - limited to windows os). Also keep in mind, Shell script (.sh) is the most common scripting language that every developer knows. Categorizing the examples has PowerShell, bash, zsh doesn't provide any value, these are simple scripting that can executed using a shell script(.sh).

For example: Let say I have a python script called app.py in my repo.

carbon

Now, I am adding the following markdown-autodocs marker in README.md.

carbon (3)

Once my changes are pushed to my branch, markdown-autodocs GitHub action will be triggered that automatically generates & updates markdown content (README.md) from external file (app.py). Then it automatically pushes, README.md changes to my branch.

carbon (2)

aspacca:
@dineshsonachalam
I realised that I prefer to maintain a single `README.md` page rather than multiple sperated files to include.
It would be different if the page would include source files that I have to duplicate the content for.
In this case anyway I would find unfortunate that I should use `bash` instead of `sh` for a bash script, for example.
I won't adopt your action, but thanks anyway for the attention.

PR #388, opened 2021-07-18
Un-hyperlink malware

tomByrer:
Don't want people to accidently click on the malware site.  :)

Looks like a neat project; bookmarked incase I'll need later!


aspacca:
thanks @tomByrer , I didn't relise that README.md rendering in github was creating a link from the text

PR #389, opened 2021-07-18
Clamav prescan

francescayeye:
- option to prescan with clamav before upload (rejecting if positive)

### DIFF
--- a/server/clamav.go
+++ b/server/clamav.go
@@ -27,18 +27,18 @@ THE SOFTWARE.
 package server
 
 import (
-	// _ "transfer.sh/app/handlers"
-	// _ "transfer.sh/app/utils"
-
+	"errors"
 	"fmt"
+	clamd "github.com/dutchcoders/go-clamd"
 	"io"
+	"io/ioutil"
 	"net/http"
 	"time"
 
-	clamd "github.com/dutchcoders/go-clamd"
-
 	"github.com/gorilla/mux"
 )
+
+const clamavScanStatusOK = "OK"
 
 func (s *Server) scanHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
@@ -50,25 +50,53 @@ func (s *Server) scanHandler(w http.ResponseWriter, r *http.Request) {
 
 	s.logger.Printf("Scanning %s %d %s", filename, contentLength, contentType)
 
-	var reader io.Reader
-
-	reader = r.Body
-
-	c := clamd.NewClamd(s.ClamAVDaemonHost)
-
-	abort := make(chan bool)
-	defer close(abort)
-	response, err := c.ScanStream(reader, abort)
+	file, err := ioutil.TempFile(s.tempPath, "clamav-")
+	defer s.cleanTmpFile(file)
 	if err != nil {
 		s.logger.Printf("%s", err.Error())
-		http.Error(w, err.Error(), 500)
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
+	_, err = io.Copy(file, r.Body)
+	if err != nil {
+		s.logger.Printf("%s", err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	status, err := s.performScan(file.Name())
+	if err != nil {
+		s.logger.Printf("%s", err.Error())
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Write([]byte(fmt.Sprintf("%v\n", status)))
+}
+
+func (s *Server) performScan(path string) (string, error) {
+	c := clamd.NewClamd(s.ClamAVDaemonHost)
+
+	responseCh := make(chan chan *clamd.ScanResult)
+	errCh := make(chan error)
+	go func(responseCh chan chan *clamd.ScanResult, errCh chan error) {
+		response, err := c.ScanFile(path)
+		if err != nil {
+			errCh <- err
+			return
+		}
+
+		responseCh <- response
+	}(responseCh, errCh)
+
 	select {
-	case s := <-response:
-		w.Write([]byte(fmt.Sprintf("%v\n", s.Status)))
+	case err := <-errCh:
+		return "", err
+	case response := <-responseCh:
+		st := <-response
+		return st.Status, nil
 	case <-time.After(time.Second * 60):
-		abort <- true
+		return "", errors.New("clamav scan timeout")
 	}
 }
--- a/server/clamav.go
+++ b/server/clamav.go
@@ -34,8 +34,8 @@ import (
 	"net/http"
 	"time"
 
-	"github.com/gorilla/mux"
 	clamd "github.com/dutchcoders/go-clamd"
+	"github.com/gorilla/mux"
 )
 
 const clamavScanStatusOK = "OK"
--- a/server/clamav.go
+++ b/server/clamav.go
@@ -29,13 +29,13 @@ package server
 import (
 	"errors"
 	"fmt"
-	clamd "github.com/dutchcoders/go-clamd"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"time"
 
 	"github.com/gorilla/mux"
+	clamd "github.com/dutchcoders/go-clamd"
 )
 
 const clamavScanStatusOK = "OK"
--- a/server/clamav.go
+++ b/server/clamav.go
@@ -72,7 +72,7 @@ func (s *Server) scanHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	w.Write([]byte(fmt.Sprintf("%v\n", status)))
+	_, _ = w.Write([]byte(fmt.Sprintf("%v\n", status)))
 }
 
 func (s *Server) performScan(path string) (string, error) {
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -291,7 +291,7 @@ func sanitize(fileName string) string {
 func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 	if err := r.ParseMultipartForm(_24K); nil != err {
 		s.logger.Printf("%s", err.Error())
-		http.Error(w, "Error occurred copying to output stream", 500)
+		http.Error(w, "Error occurred copying to output stream", http.StatusInternalServerError)
 		return
 	}
 
@@ -309,7 +309,7 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 
 			if f, err = fheader.Open(); err != nil {
 				s.logger.Printf("%s", err.Error())
-				http.Error(w, err.Error(), 500)
+				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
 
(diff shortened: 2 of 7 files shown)


server/handlers.go
```diff
@@ -299,63 +311,85 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
...
+				s.logger.Print("Entity too large")
 				http.Error(w, http.StatusText(http.StatusRequestEntityTooLarge), http.StatusRequestEntityTooLarge)
 				return
 			}
 
+			if s.performClamavPrescan {
+				status, err := s.performScan(reader)
+				if err != nil {
+					s.logger.Printf("%s", err.Error())
+					http.Error(w, errors.New("Could not perform prescan").Error(), 500)
```
stefanbenten, line 361: Just to be sane, lets do http.StatusInternalServerError here?

server/handlers.go
```diff
@@ -299,63 +311,85 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
...
 
 			if s.maxUploadSize > 0 && contentLength > s.maxUploadSize {
-				log.Print("Entity too large")
+				s.logger.Print("Entity too large")
 				http.Error(w, http.StatusText(http.StatusRequestEntityTooLarge), http.StatusRequestEntityTooLarge)
 				return
 			}
 
+			if s.performClamavPrescan {
+				status, err := s.performScan(reader)
```
stefanbenten, line 358: If the scan is performed using the reader directly, we should not be able to actually upload the data to the backend storage since the following reader in L391 would directly return a EOF, right?

server/handlers.go
```diff
@@ -430,9 +464,15 @@ func (s *Server) putHandler(w http.ResponseWriter, r *http.Request) {
...
 	contentLength := r.ContentLength
 
-	var reader io.Reader
+	var reader io.ReadSeeker
 
-	reader = r.Body
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		s.logger.Printf("%s", err.Error())
+		http.Error(w, err.Error(), 500)
```
stefanbenten, line 472: http.StatusInternalServerError

server/handlers.go
```diff
@@ -447,79 +487,96 @@ func (s *Server) putHandler(w http.ResponseWriter, r *http.Request) {
...
 		}
 
 		contentLength = n
 	}
 
+	if s.performClamavPrescan {
+		status, err := s.performScan(reader)
+		if err != nil {
+			s.logger.Printf("%s", err.Error())
+			http.Error(w, errors.New("Could not perform prescan").Error(), 500)
```
stefanbenten, line 530: http.StatusInternalServerError

stefanbenten (commented):
I saw we have used the int status code in various places, so not the biggest need to change those.
But from what i can see the actual scanning would void the data and the actual backend upload should not get any data.

server/handlers.go
aspacca: it's a readseeker, it is rewind here https://github.com/dutchcoders/transfer.sh/pull/389/files#diff-c93b35289b4afc168cec4f328144c11c310ac6a39f4495bdf8a9bf4c82513f40R77

stefanbenten: Oh that makes sense.

server/clamav.go
```diff
@@ -50,26 +53,44 @@ func (s *Server) scanHandler(w http.ResponseWriter, r *http.Request) {
 
 	s.logger.Printf("Scanning %s %d %s", filename, contentLength, contentType)
 
-	var reader io.Reader
+	var reader io.ReadSeeker
 
-	reader = r.Body
+	body, err := ioutil.ReadAll(r.Body)
```
stefanbenten, line 58: This can cause OOM's very quickly with the nature that we allow 10G files.
Having two 10G files entirely read like this should blocks 20G of memory.

I guess there is not much other we can do than writing it into a temp file first to avoid bloating the memory..

server/handlers.go
```diff
@@ -442,9 +464,15 @@ func (s *Server) putHandler(w http.ResponseWriter, r *http.Request) {
 
 	contentLength := r.ContentLength
 
-	var reader io.Reader
+	var reader io.ReadSeeker
 
-	reader = r.Body
+	body, err := ioutil.ReadAll(r.Body)
```
stefanbenten, line 469: same as above.

stefanbenten:
I'll review/test this tomorrow 👍

server/handlers.go
```diff
@@ -449,53 +459,54 @@ func (s *Server) putHandler(w http.ResponseWriter, r *http.Request) {
...
-		// queue file to disk, because s3 needs content length
-		var err error
-		var f io.Reader
-
-		f = reader
-
-		var b bytes.Buffer
+	var err error
+	var file *os.File
+	defer s.cleanTmpFile(file)
```
stefanbenten, line 466: Lets move this after the actual creation/assignment in L468. On my tests the files didnt get removed for both failures and successes.

server/server.go
```diff
@@ -53,14 +53,12 @@ import (
 	"crypto/tls"
 
 	web "github.com/dutchcoders/transfer.sh-web"
-	assetfs "github.com/elazarl/go-bindata-assetfs"
 
-	autocert "golang.org/x/crypto/acme/autocert"
+	"github.com/elazarl/go-bindata-assetfs"
+	"golang.org/x/crypto/acme/autocert"
```
stefanbenten, line 58: Imports are not sorted properly here.

stefanbenten (commented):
Overall worked fine for me, just a couple of small changes.

stefanbenten:
For bigger files the upload also seems to "hang" while the prescan is ongoing. Should we add something to the web interface that shows this "step"?

aspacca:
>For bigger files the upload also seems to "hang" while the prescan is ongoing. Should we add something to the web interface that shows this "step"?

I haven't tested on UI, I will check.

The main problem is due to clamd limits:
1- I had to change from stream scan to file scan, on big files it would take forever with the stream
2- Maximum scan dimension is 4TB, and is still advised by clamav documentation to not set so high. Bigger file will be reported as clean. We should document this and probably apply automatically `max-upload-size` to match the max scan size set in clamd if clam prescan is enabled.

server/handlers.go
aspacca: good catch!

stefanbenten:
I'll configured clamd for 4G files. Something i was thinking of was is splitting the temp files. I need to test if that works for a single binary thats malicious getting split up into say 4 smaller files and then scanned in parallel. In practice i would assume finding the same matches, but statistically, i might split the malicious code apart not letting clamd recognize it.

I wonder if for the UI we can simply adjust the progress bar.

aspacca:
I think that splitting the file could lead to false negative.
In UI I'm not sure how we can handle: progress bar is rendered according to the upload progress

stefanbenten:
I need to dive into that frontend code to look further, but yeah unlikely an easy way.

stefanbenten:
@aspacca Lets get this pushed along and limit file size scan limit to something like 1G for now. It is better than nothing 👍

aspacca:
@stefanbenten I wanted to refactor this PR in order to allow both `ScanStream` and `ScanFile` according to the type of clamav property (file or stream)

also setting automatically the file limit in transfer.sh if the prescan is enabled

aspacca:
@stefanbenten feel free to merge after testing

stefanbenten:
Thanks, will push the linting fixes as well after testing!

aspacca:
I did the linting

aspacca:
hi @stefanbenten 

for me it is good to merge, I will later refine something around `CLAMAV_HOST` (now it will only work with local clamav instance)

stefanbenten:
Thanks for pushing this up. I'll give it a go in a bit :+1:

stefanbenten:
Working like a charm! Thanks again @aspacca !

aspacca:
let's wait to make a release when I'll add back support for clamav on the network :)

LinusCDE:
This change along with missing (`/tmp`) broke my setup today. I'm using watchtower to update daily and today my transfer.sh instance went down due to this error:

grafik

I could trace this back to the [newest commit](https://github.com/dutchcoders/transfer.sh/commit/cff0a88bf3d90f4de654dcdb1157ce0b2bccf02c#diff-8e494a434a8037b6c0b888e25b2baae7618fe65e792d4a155dadd096e9350667R396) and this PR. It seems that setting either `CLAMAV_HOST` to anything other than an empty string or `PERFORM_CLAMAV_PRESCAN` to `true` solved the issue. Having the prescan set to true while it doesn't affect the runtime without any clamav seems weird though. Also sure the panic should happen if `CLAMAV_HOST` is not set AND ONLY when `PERFORM_CLAMAV_PRESCAN` is false??

Aside: While looking up the command arguments, I also noticed that, on the [dockerhub site](https://hub.docker.com/r/dutchcoders/transfer.sh), `provider` has no env specified while it does exist in the go code. So that could probably be added to dockerhub as well.

My setup is pretty simple btw: Just a local file provider with a mounted volume for the files. Also needed to add a volume `/tmp` now as well or any uploads would fail with 500 for missing "/tmp/transfer-something`.

aspacca:
@LinusCDE unfortunately we cannot garantuee issues like this to happen on main branch and docker images based on main branch. You must consider main as unstable

The problem is here: https://github.com/dutchcoders/transfer.sh/pull/459/files#diff-8e494a434a8037b6c0b888e25b2baae7618fe65e792d4a155dadd096e9350667L396: `!v` shoud be `v`

fixed on https://github.com/dutchcoders/transfer.sh/pull/460

LinusCDE:
> @LinusCDE unfortunately we cannot garantuee issues like this to happen on main branch and docker images based on main branch. You must consider main as unstable

I can understand this way of thinking. However I intuitively though the other way. Since some consider latest/main/master stable (having a dev branch or unstable builds) it can usually either be "stable" or "unstable". Seeing that on dockerhub, there are two "unstable" tags "unstable" / "edge" and no explicit latest stable ones like "stable" "v1" or "1", I assumed latest to be the stable one. Even though latest gets updated daily, I just though this would be because the build is scheduled daily.

Should I create an issue as well for this?

Since I have auto-updating and there is no tag for the latest stable release, this would force me update the tag manually for each release and at sometime forget about this and let the version rot. I use [watchtower](https://containrrr.dev/watchtower/) to keep my containers/services up-to-date and usually try to make it target the latest stable branch (preferably of the current major version like "1", so I still need to address breaking changes manually).

MarkusFreitag:
@aspacca Do you mean that there is no tag that can be used to pull the current stable release?
Honestly I considered `edge` to be the unstable build and `latest` to be a link to the current release.

stefanbenten:
We probably could make a `stable` tag and only update it when we push tags and its corresponding builds. How would that sound?

LinusCDE:
That sounds great!

aspacca:
> @aspacca Do you mean that there is no tag that can be used to pull the current stable release? Honestly I considered `edge` to be the unstable build and `latest` to be a link to the current release.

`latest` is an alias to `edge`/`nightly`. `edge` is built on the latest commit on master, `nightly` is schedelude every day (https://github.com/dutchcoders/transfer.sh/blame/main/.github/workflows/build-docker-images.yml#L38-L40)

In the end `edge` and `nightly` will converge, but until the next daily schedule `edge` will be forward to `nightly`

I personally agree with https://vsupalov.com/docker-latest-tag/, as long as we talk about production environment.
I think it's should be up to any CD logic to fetch/discover the most updated versioned tag and deploy it.

Anyway, let's add a `stable` tag :)

LinusCDE:
On production, where people are paid to monitor their deployment this makes sense. But not when I'm having dozens of containers, that either update automatically or get forgotten.

Also production seems to usually have a host of other problems (namely having own up to date builds and not waiting on a dependency chain to fix CVEs). So they will most likely create their own containers anyway, and this is what it seems to target.

PR #395, opened 2021-08-11
Build more binary releases

wc7086:


.github/workflows/release.yml
```diff
@@ -1,59 +1,174 @@
...
+        run: |
+          cd ./build_assets || exit 1
+          mv transfersh transfersh.exe
+
+      - name: Create ZIP archive
+        shell: bash
+        run: |
+          pushd build_assets || exit 1
+          touch -mt $(date +%Y01010000) *
+          zip -9vr ../Transfersh-$ASSET_NAME.zip .
```
aspacca, line 147: I'd keep `transfersh` lowercase and the `${GITHUB_REF##*/}` suffix

aspacca: I would prefer tar gz archive as well

.github/workflows/release.yml
```diff
@@ -1,59 +1,174 @@
...
+name: Build and Release
+
 on:
+  workflow_dispatch:
+  release:
+    types: [published]
   push:
-    tags:
-      - v*
+    branches:
```
aspacca, line 8: what's the goal of building binaries on master and PRs?
be aware of build failures?

aspacca (changes requested):

aspacca:
@wc7086 thanks for the contribution, please see my comments

.github/workflows/release.yml
```diff
@@ -1,59 +1,174 @@
...
-          name: artifacts
-          path: ./artifacts
-      - name: Publish artifacts
-        uses: softprops/action-gh-release@v1
+          name: Transfersh-${{ steps.get_filename.outputs.ASSET_NAME }}
+          path: |
+            ./Transfersh-${{ steps.get_filename.outputs.ASSET_NAME }}/*
+
+      - name: Upload binaries to release
+        uses: svenstaro/upload-release-action@v2
```
aspacca, line 168: `svenstaro/upload-release-action@v2` doesn't seem to be actively maintained, I'd stick to `softprops/action-gh-release@v1` if possible

.github/workflows/release.yml
```diff
@@ -1,59 +1,174 @@
...
+        uses: actions/setup-go@v2
         with:
-          name: artifacts
-          path: ./artifacts
+          go-version: ^1.16
+ 
+      - name: Build Transfersh
+        run: |
+          mkdir -p build_assets
+          go build -tags netgo -ldflags "-X github.com/dutchcoders/transfer.sh/cmd.Version=${GITHUB_REF##*/} -a -s -w -extldflags '-static'" -o build_assets/transfersh
```
aspacca, line 129: I would keep the binary filename with the os/arch suffix, you could use the friendly name as well

.github/workflows/release.yml
aspacca: actually this is going to change the format of the assets filename (as well the content), potentially breaking automated tools.
I would keep not compressed artifacts of the binary, with the same naming format we already have. please be sure that for existing architectures and OSes the friendly name will not change what we already have

aspacca (changes requested):

.github/workflows/release.yml
```diff
@@ -1,59 +1,174 @@
...
+        run: |
+          pushd build_assets || exit 1
+          touch -mt $(date +%Y01010000) *
+          zip -9vr ../Transfersh-$ASSET_NAME.zip .
+          popd || exit 1
+          FILE=./Transfersh-$ASSET_NAME.zip
+          DGST=$FILE.dgst
+          for METHOD in {"md5","sha1","sha256","sha512"}
+          do
+            openssl dgst -$METHOD $FILE | sed 's/([^)]*)//g' >> $DGST
```
aspacca, line 153: I would rather generate separated digest files

aspacca (changes requested):

.github/workflows/release.yml
wc7086: yes

wc7086:
This is modified from the automated yaml of other projects. Because I don’t know everyone’s preferences, and I didn’t consider some automated programs, so I didn’t write very well. It has now been modified according to your requirements.

.github/workflows/release.yml
wc7086: But I don't know much about golang, so I don't know if it is necessary to test the binary file build. What do you think?

wc7086:
I made some mistakes which led to more commits. Should I open another PR or you use the squash mode to merge my commits?

wc7086:
I made some mistakes which led to more commits. Should I open another PR or you use the squash mode to merge my commits?

stefanbenten:
We can squash them 👍

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
+{
+    "android-arm64": { "friendlyName": "android-arm64-v8a" },
+    "darwin-amd64": { "friendlyName": "macos-64" },
```
aspacca, line 3: this should be `darwin-amd64` in order to not change the format from previous release

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "darwin-arm64": { "friendlyName": "macos-arm64-v8a" },
```
aspacca, line 4: this should be `darwin-arm64` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "freebsd-386": { "friendlyName": "freebsd-32" },
+    "freebsd-amd64": { "friendlyName": "freebsd-64" },
+    "freebsd-arm64": { "friendlyName": "freebsd-arm64-v8a" },
+    "freebsd-arm7": { "friendlyName": "freebsd-arm32-v7a" },
+    "linux-386": { "friendlyName": "linux-32" },
+    "linux-amd64": { "friendlyName": "linux-64" },
+    "linux-arm5": { "friendlyName": "linux-arm32-v5" },
+    "linux-arm64": { "friendlyName": "linux-arm64-v8a" },
+    "linux-arm6": { "friendlyName": "linux-arm32-v6" },
+    "linux-arm7": { "friendlyName": "linux-arm32-v7a" },
```
aspacca, line 15: this should be `linux-armv7` in order to not change the format from previous release

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "linux-ppc64le": { "friendlyName": "linux-ppc64le" },
+    "linux-ppc64": { "friendlyName": "linux-ppc64" },
+    "linux-riscv64": { "friendlyName": "linux-riscv64" },
+    "linux-s390x": { "friendlyName": "linux-s390x" },
+    "openbsd-386": { "friendlyName": "openbsd-32" },
+    "openbsd-amd64": { "friendlyName": "openbsd-64" },
+    "openbsd-arm64": { "friendlyName": "openbsd-arm64-v8a" },
+    "openbsd-arm7": { "friendlyName": "openbsd-arm32-v7a" },
+    "windows-386": { "friendlyName": "windows-32" },
+    "windows-amd64": { "friendlyName": "windows-64" },
```
aspacca, line 31: this should be `windows-amd64` in order to not change the format from previous release

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
+{
+    "android-arm64": { "friendlyName": "android-arm64-v8a" },
+    "darwin-amd64": { "friendlyName": "macos-64" },
+    "darwin-arm64": { "friendlyName": "macos-arm64-v8a" },
+    "dragonfly-amd64": { "friendlyName": "dragonfly-64" },
```
aspacca, line 5: this should be `dragonfly-arm64` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "freebsd-386": { "friendlyName": "freebsd-32" },
```
aspacca, line 6: this should be `freebsd-386` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "freebsd-amd64": { "friendlyName": "freebsd-64" },
```
aspacca, line 7: this should be `freebsd-amd64` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "freebsd-arm64": { "friendlyName": "freebsd-arm64-v8a" },
+    "freebsd-arm7": { "friendlyName": "freebsd-arm32-v7a" },
+    "linux-386": { "friendlyName": "linux-32" },
```
aspacca, line 10: this should be `linux-386` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "linux-mipslesoftfloat": { "friendlyName": "linux-mips32le-softfloat" },
+    "linux-mipsle": { "friendlyName": "linux-mips32le" },
+    "linux-mipssoftfloat": { "friendlyName": "linux-mips32-softfloat" },
+    "linux-mips": { "friendlyName": "linux-mips32" },
+    "linux-ppc64le": { "friendlyName": "linux-ppc64le" },
+    "linux-ppc64": { "friendlyName": "linux-ppc64" },
+    "linux-riscv64": { "friendlyName": "linux-riscv64" },
+    "linux-s390x": { "friendlyName": "linux-s390x" },
+    "openbsd-386": { "friendlyName": "openbsd-32" },
+    "openbsd-amd64": { "friendlyName": "openbsd-64" },
```
aspacca, line 27: this should be `openbsd-amd64` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "linux-mips64": { "friendlyName": "linux-mips64" },
+    "linux-mipslesoftfloat": { "friendlyName": "linux-mips32le-softfloat" },
+    "linux-mipsle": { "friendlyName": "linux-mips32le" },
+    "linux-mipssoftfloat": { "friendlyName": "linux-mips32-softfloat" },
+    "linux-mips": { "friendlyName": "linux-mips32" },
+    "linux-ppc64le": { "friendlyName": "linux-ppc64le" },
+    "linux-ppc64": { "friendlyName": "linux-ppc64" },
+    "linux-riscv64": { "friendlyName": "linux-riscv64" },
+    "linux-s390x": { "friendlyName": "linux-s390x" },
+    "openbsd-386": { "friendlyName": "openbsd-32" },
```
aspacca, line 26: this should be `openbsd-386` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "openbsd-amd64": { "friendlyName": "openbsd-64" },
+    "openbsd-arm64": { "friendlyName": "openbsd-arm64-v8a" },
+    "openbsd-arm7": { "friendlyName": "openbsd-arm32-v7a" },
+    "windows-386": { "friendlyName": "windows-32" },
```
aspacca, line 30: this should be `windows-386` or similar for coherence with the other one

.github/build/friendly-filenames.json
```diff
@@ -0,0 +1,33 @@
...
+    "android-arm64": { "friendlyName": "android-arm64-v8a" },
+    "darwin-amd64": { "friendlyName": "macos-64" },
+    "darwin-arm64": { "friendlyName": "macos-arm64-v8a" },
+    "dragonfly-amd64": { "friendlyName": "dragonfly-64" },
+    "freebsd-386": { "friendlyName": "freebsd-32" },
+    "freebsd-amd64": { "friendlyName": "freebsd-64" },
+    "freebsd-arm64": { "friendlyName": "freebsd-arm64-v8a" },
+    "freebsd-arm7": { "friendlyName": "freebsd-arm32-v7a" },
+    "linux-386": { "friendlyName": "linux-32" },
+    "linux-amd64": { "friendlyName": "linux-64" },
```
aspacca, line 11: this should be `linux-amd64` in order to not change the format from previous release

.github/workflows/release.yml
aspacca: it depends on the tests coverage: ideally if tests run then it should also compile

@stefanbenten what do you think?

.github/workflows/release.yml
```diff
@@ -1,59 +1,181 @@
...
+          name: transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}
+          path: |
+            ./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}/*
+
+      - name: Upload binaries to release
         uses: softprops/action-gh-release@v1
+        if: github.event_name == 'release'
         with:
-          files: './artifacts/*'
+          files: ./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}.tar.gz*
```
aspacca, line 178: please, release both gz archive and bare binary. you could add a sha256 digest for them as well

aspacca (changes requested):

wc7086:
In order to facilitate the tracking of the problem, I think that the version number should be written into the cmd.go and the Commit ID should be added. What do you think?

.github/workflows/release.yml
```diff
@@ -1,59 +1,186 @@
...
+          path: |
+            ./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}/*
+
+      - name: Upload binaries to release
         uses: softprops/action-gh-release@v1
+        if: github.event_name == 'release'
         with:
-          files: './artifacts/*'
+          files: |
+            ./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}.tar.gz*
```
aspacca, line 182: this glob it's not needed: the following one will catch also `.tar.gz*`

aspacca (changes requested):

.github/workflows/release.yml
wc7086: Okay. I should remove https://github.com/wc7086/transfer.sh/blob/f63c8c34685875e1a20bf5bf73db2c561dfb6962/.github/workflows/release.yml#L157.

wc7086: If don't move `tar.gz` to the upper directory, will need more steps to process the content.

.github/workflows/release.yml
stefanbenten: I do not think we need to build the assets for PRs, since the tests should cover the actual compile yeah.
For master its debatable if users would like to get the binaries for non-releases, generally should not be a big deal tho.

aspacca: @wc7086 please remove then, thanks

aspacca:
>In order to facilitate the tracking of the problem, I think that the version number should be written into the cmd.go and the Commit ID should be added. What do you think?

that's what happens here: https://github.com/dutchcoders/transfer.sh/pull/395/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R126

wc7086:
>that's what happens here: https://github.com/dutchcoders/transfer.sh/pull/395/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R126

This is the tag, not the Commit ID.
I circled the Commit ID in the image below.
image

aspacca:
> This is the tag, not the Commit ID.

the tag points to a commit, if you build from a not tagged commit it will print something like `v1.x.x-commit`

.github/workflows/release.yml
```diff
@@ -1,59 +1,179 @@
...
+            ./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}/./transfersh-${{ steps.get_filename.outputs.GIT_TAG }}-${{ steps.get_filename.outputs.ASSET_NAME }}*
```
aspacca, line 176: are you sure about this path?

wc7086: I don't know why I hit one `.`, maybe it was too sleepy at the time. But I didn't notice it because it worked well...
https://github.com/wc7086/transfer.sh/releases/tag/v1.3.3
already edited.

PR #399, opened 2021-08-15
s3 and storj storage: use nill when purgeDays is 0

### LINKED ISSUE #398: Using storj storage without setting purge-days would cause upload fail
https://github.com/dutchcoders/transfer.sh/blob/d16a6cc25e14468db3cdb8ceca8fad0b792dec79/server/storage.go#L716

The default value of purge-days is 0. This causes the uplink to trigger an error (Invalid expiration time).
purge-days的預設值為0。會導致uplink觸發錯誤(Invalid expiration time)。

擷取
擷取2

francescayeye:
fixes #398

### DIFF
--- a/server/storage.go
+++ b/server/storage.go
@@ -273,11 +273,16 @@ func (s *S3Storage) Put(token string, filename string, reader io.Reader, content
 		u.LeavePartsOnError = false
 	})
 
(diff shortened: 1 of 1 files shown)


stefanbenten (approved):
Works 👍

PR #418, opened 2021-10-09
Docker: Allow selection of (unprivileged) UID/GID at build time

lkubb:
I created a pull request to discuss my proposal in a dedicated thread and allow merging, if desired. I hope that's fine and I'm doing everything correctly, as I said, I'm mostly a lone lurker on here. Previous discussion is in https://github.com/dutchcoders/transfer.sh/pull/345, which is related. I made the newbie mistake and commited to main, so I had to clean up a little, which is why the previous links are a bit borked.

This PR introduces the option to create a Docker container running as an unprivileged user while retaining the previous default of running as root, and while still being based on `FROM scratch`. This avoids a breaking change for existing users and keeps the image as lean as it is. UID/GID can optionally be specified as well.

Setting the build argument `RUNAS` to anything will trigger the unprivileged mode. If it's unset, Docker will happily default to root in both `COPY --chown=${RUNAS}` and `USER ${RUNAS}`. To avoid an error when the glob `/tmp/useradd/*` returns nothing, I had to create a dummy file when building for root. The (relatively speaking) resource-heavy steps of compiling can be reused from cache when building different variants of the image.

* `docker build -t transfersh-root .` will result in an image which runs the container as root. The only change from before is the dummy file `/etc/unused`.
* `docker build -t transfersh-user --build-arg RUNAS=plsanythingbutroot .` will result in an image that runs the container with the default UID:GID of 5000:5000.
* `docker build -t transfersh-user-custom --build-arg RUNAS=verycustomsir --build-arg PUID=1337 --build-arg PGID=1338 .` will result in an image that runs the container with UID:GID of 1337:1338.

Apart from the good conscience of running as an unprivileged process, this also allows easy selection of an ID, e.g. when mapping NFS-mounted folders as the base dir. Furthermore, it would be possible to provide a separate official image like `transfersh:noroot` without much hassle. Current users wanting to switch manually would have to run a simple `sudo chown -R $UID:$GID ./data` on their host.

Fixes https://github.com/dutchcoders/transfer.sh/issues/100.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ You can also build the container yourself. This allows you to choose which UID/G
 # * PUID:  UID of the process. Needs RUNAS != "". Defaults to 5000.
 # * PGID:  GID of the process. Needs RUNAS != "". Defaults to 5000.
 
-docker build -t tfsh-noroot --build-arg RUNAS=doesntmatter --build-arg PUID=1337 --build-arg PGID=1338 .
+docker build -t transfer.sh-noroot --build-arg RUNAS=doesntmatter --build-arg PUID=1337 --build-arg PGID=1338 .
 ```
 
 ## S3 Usage
--- a/README.md
+++ b/README.md
@@ -140,10 +140,29 @@ $ go build -o transfersh main.go
 
 ## Docker
 
-For easy deployment, we've created a Docker container.
+For easy deployment, we've created an official Docker container. There are two variants, differing only by which user runs the process.
+
+The default one will run as `root`:
 
 ```bash
 docker run --publish 8080:8080 dutchcoders/transfer.sh:latest --provider local --basedir /tmp/
+```
+
+The one tagged with the suffix `-noroot` will use `5000` as both UID and GID:
+```bash
+docker run --publish 8080:8080 dutchcoders/transfer.sh:latest-noroot --provider local --basedir /tmp/
+```
+
+### Building the Container
+You can also build the container yourself. This allows you to choose which UID/GID will be used, e.g. when using NFS mounts:
+```bash
+# Build arguments:
+# * RUNAS: If empty, the container will run as root.
+#          Set this to anything to enable UID/GID selection.
+# * PUID:  UID of the process. Needs RUNAS != "". Defaults to 5000.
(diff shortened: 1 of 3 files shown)


aspacca:
@lkubb , great!

are you able to update the documentation as well: https://github.com/dutchcoders/transfer.sh#docker? (adding an how to build step)

the best would be to add as well another github workflow (https://github.com/dutchcoders/transfer.sh/blob/main/.github/workflows/build-docker-images.yml) in order to build the non-root image :)

uptimejeff:
Thanks for your work on this... running as root didn't work for me either :-)

Original build command from directions
`docker build -t transfersh-user --build-args RUNAS=plsanythingbutroot .`
Which returns
`unknown flag: --build-args`

Switched to arg (singular) and ran
`docker build -t transfersh-user --build-args RUNAS=plsanythingbutroot .`
The build ends here
```
Step 11/18 : LABEL maintainer="Andrea Spacca <andrea.spacca@gmail.com>"
unexpected EOF
```

README.md
```diff
@@ -140,12 +140,31 @@ $ go build -o transfersh main.go
...
+### Building the Container
+You can also build the container yourself. This allows you to choose which UID/GID will be used, e.g. when using NFS mounts:
+```bash
+# Build arguments:
+# * RUNAS: If empty, the container will run as root.
+#          Set this to anything to enable UID/GID selection.
+# * PUID:  UID of the process. Needs RUNAS != "". Defaults to 5000.
+# * PGID:  GID of the process. Needs RUNAS != "". Defaults to 5000.
+
+docker build -t tfsh-noroot --build-arg RUNAS=doesntmatter --build-arg PUID=1337 --build-arg PGID=1338 .
```
aspacca, line 165: minor: I would change `tfsh-noroot` to `transfer.sh-noroot`, to increase readability :)

aspacca (approved):

aspacca:
hi @lkubb 

thank you very much and I apologise as well for taking a little time to review.

I would just ask you to address the comment I left: I will wait a couple of days and then merge anyway and "fix" on my own :)

lkubb:
No worries, I'm glad I'm in the position to contribute and you're happy with the result. :) Just addressed your comment and rebased again on main. Cheers!

Issue #420, opened 2021-10-12
No error returned when Max-Days is invalid


anihm136:
When a file is uploaded with a very high value set for `Max-Days` header, the server returns a URL. However, the URL leads to a 404 page

**Expected behavior**:
The server should return an appropriate 400 response when an invalid value is set for the header

kugiyasan:
Hey can I take this issue?

aspacca:
@anihm136 how much high the value must be set?

anihm136:
Maybe overkill, but I did a binary search and it appears to fail at 8217934 days

jimmybrancaccio:
Looks good now @aspacca and @kugiyasan Thank you both! ❤️

PR #422, opened 2021-10-16
issue #420 return 400 response when Max-Days is too big

kugi83:


aspacca:
@kugiyasan thanks for trying to solve the issue, but this doesn't seem the proper solution.
I still have to understand what's exactly the bug: and integer overflow or a problem in the check for the max date constraint.

your code doesn't fix either of them: also the metadata are saved anyway even if the max date is "wrong"

kugiyasan:
Oh sorry, I jumped the gun and forgot to explain!

I played around with a local server, and `Max-Days: 100000` is valid, while `Max-Days: 110000` is invalid. Looking at the source code, this header is converted to an `int` and is then passed to `time.Duration`. `time.Duration`'s documentation says that: 

> The representation limits the largest representable duration to approximately 290 years.

which explains why 100000 days (around 274 years) works but 110000 days (301 years) fails.

Personally I prefer the 400 response. but both numbers aren't logically speaking "invalid", so should we simply silently ignore `Max-Days` bigger than 106752 days so that the uploaded file acts like it has no expired date?

PR #424, opened 2021-10-22
Implement Nix Flake

ysndr:
This PR makes transfer.sh available to the nix works by providing a nixos Module and standalone executable (on non-nixos machines) as well as a default configuration and basic container image definition.﻿

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -102,7 +102,7 @@
                   basedir = mkOption { type = types.str; description = "path storage for local provider"; default = "${cfg.stateDir}/store"; };
                   purge-interval = mkOption { type = types.nullOr types.int; description = "interval in hours to run the automatic purge for (not applicable to S3 and Storj)"; };
                 };
-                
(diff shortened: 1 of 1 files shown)


aspacca:
@ysndr 
I bumped a dependency and I have a doubt about https://github.com/dutchcoders/transfer.sh/blob/main/flake.nix#L11
From nix documentation:
```
vendorSha256: is the hash of the output of the intermediate fetcher derivation. vendorSha256 can also take null as an input. When null is used as a value, rather than fetching the dependencies and vendoring them, we use the vendoring included within the source repo. If you’d like to not have to update this field on dependency changes, run go mod vendor in your source repo and set vendorSha256 = null;
```

If I got it right it's the sh256 of the vendor directory, bumping the version of one dependency should have changed it, so the value should be updated.

What's the cons of setting `vendorSha256 = null` and run go mod vendor?

PR #432, opened 2021-11-01
Run app with non-root user; fix #100

### LINKED ISSUE #100: [Security Issue] Docker containers' processes run as root

BOPOHA:


aspacca:
@BOPOHA we already have https://github.com/dutchcoders/transfer.sh/pull/418 for this, sorry if you spent time on the issue

PR #433, opened 2021-11-01
added s3-ec2role provider; fix #431

### LINKED ISSUE #431: Feature: S3 Usage with an AWS IAM Role
To automate the deployment process in a aws cloud, it would be great to use the AWS IAM Role to access the s3 bucket.

BOPOHA:


cmd/cmd.go
```diff
@@ -445,6 +445,14 @@ func New() *Cmd {
 			} else {
 				options = append(options, server.UseStorage(storage))
 			}
+		case "s3-ec2role": // using AWS IAM Role and EC2 Profile
```
aspacca, line 448: this is not really needed.

you should rather add something like `s3-credentials-type` config param, default to `access-key` and refactor https://github.com/dutchcoders/transfer.sh/blob/e97fdcb2930a3d2d139c80ec1cd3caf521d50b7b/server/storage.go#L176 and https://github.com/dutchcoders/transfer.sh/blob/main/server/utils.go#L43

francescayeye (changes requested):
see my comment, thanks

stefanbenten:
@BOPOHA Do you intend to address the comment from @francescayeye ?

BOPOHA:
@stefanbenten  thanks for following up.

I revisited this PR against the current codebase and updated the implementation locally. Since the original proposal, transfer.sh has migrated to AWS SDK for Go v2, which provides a comprehensive default credential chain.

Based on the original review feedback, I agree that a separate `s3-ec2role` provider is no longer the right abstraction. The storage provider remains `s3`; only credential discovery changes.

The updated approach is:

  - Keep the existing `s3` provider.
  - Preserve static credentials when both `aws-access-key` and `aws-secret-key` are configured.
  - Use the AWS SDK default credential chain when neither is configured.
  - Reject configurations where only one static credential is supplied.
  - Add tests and documentation covering static and role-based credentials.

This enables EC2 instance profiles, ECS task roles, EKS IRSA, standard AWS environment variables, and shared AWS configuration while remaining backward-compatible.

I have also updated #431 with the current motivation and acceptance criteria. I’m preparing to replace the old implementation in this PR with the revised version.

cc @francescayeye

k8s/transfer.sh/values.yaml
```diff
@@ -45,7 +45,8 @@ transfersh:
     pathStyle: false
     # Set to true to disable multipart uploads
     noMultipart: false
-    # Use an existing secret for AWS credentials
+    # Use an existing secret for static AWS credentials. If no credentials are set,
+    # the AWS SDK default credential chain is used (for example, EC2 roles or EKS IRSA).
```
francescayeye: i still honestly don't like this defaulting behaviour because it's a breaking change on current configuration when both `aws-access-key` and `aws-secret-key` are forgot empty.
i still suggest the introduction of a `s3-credentials-type` config that defaults to the legacy behaviour and if explicitly set to "default-sdk-credential-chain" (or whatever better value we will find for it) will activate the new feature.
@stefanbenten what do you think about this?
we will need to bump a minor for this feature to be released

README.md
```diff
@@ -334,11 +334,15 @@ docker build -t transfer.sh-noroot --build-arg RUNAS=doesntmatter --build-arg PU
 
 For the usage with a AWS S3 Bucket, you just need to specify the following options:
 - provider `--provider s3`
-- aws-access-key _(either via flag or environment variable `AWS_ACCESS_KEY`)_
-- aws-secret-key _(either via flag or environment variable `AWS_SECRET_KEY`)_
 - bucket _(either via flag or environment variable `BUCKET`)_
 - s3-region _(either via flag or environment variable `S3_REGION`)_
 
+Authentication uses the AWS SDK default credential chain. This supports sources such as
```
francescayeye: this is a change of the standard formatting for every configuration flags on this section of the README. we will have until we have an agreement on my proposal of a `s3-credentials-type` new configuration flags, i would say to leave it open to properly reformatting this bit until we will get there

BOPOHA:
Thanks for the feedback. I implemented the requested explicit credential mode:

- Added --s3-credentials-type and S3_CREDENTIALS_TYPE.
- The default value is legacy, preserving the existing requirement for both aws-access-key and aws-secret-key.
- Setting the value to default-sdk-credential-chain explicitly enables the AWS SDK credential chain for environment credentials, shared configuration, ECS task roles, EC2 instance profiles, and EKS IRSA.
- Unsupported credential types now return a clear error.
- The Helm chart exposes transfersh.s3.credentialsType with legacy as its default and only creates or mounts static credential Secrets in legacy mode.
- Added tests for the default, legacy validation, SDK-chain selection, unsupported values, and AWS session-token handling.

Validation completed successfully with go test, go vet, Helm lint, and rendered manifests for both credential modes.

PR #435, opened 2021-11-11
Add X-Url-Delete-* headers to POST handler

### LINKED ISSUE #419: No `x-url-delete` header when sending POST request
I noticed that the `x-url-delete` header is not being set when the user uploads a file via post request. Is uploading files via POST deprecated? (since all the examples seem to use PUT requests instead). If not, I think the header should be set correctly

anihm136:
Fixes #419 
Adds X-Url-Delete headers to the post handler. Since the post handler currently supports multiple file uploads (as multipart file form), currently the header is postfixed with the token of the file to create a unique delete token for each file (any better suggestions for handling this can be implemented)


aspacca (approved):

aspacca:
> any better suggestions for handling this can be implemented

We could send multiple `X-Url-Delete` headers, with every delete url.
Replacing
`w.Header().Set("X-Url-Delete-"+token, resolveURL(r, deleteURL, s.proxyPort))`
with 
`w.Header().Add("X-Url-Delete", resolveURL(r, deleteURL, s.proxyPort))`
should be enough

The delete url already contains the full get url including the filename `https://transfer.sh/QMXz5o/hello.txt/e199GfHtnsyC` so users should be able to match

server/handlers.go
```diff
@@ -389,6 +389,8 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 
 			filename = url.PathEscape(filename)
 			relativeURL, _ := url.Parse(path.Join(s.proxyPath, token, filename))
+			deleteURL, _ := url.Parse(path.Join(s.proxyPath, token, filename, metadata.DeletionToken))
+			w.Header().Set("X-Url-Delete-"+token, resolveURL(r, deleteURL, s.proxyPort))
```
aspacca, line 393: you cannot send headers after sending body content below

aspacca (changes requested):

server/handlers.go
anihm136: I did not understand this. Do you mean to say that we cannot set headers after writing to `w` using `Fprintf`? If so, then how does it work to handle multiple files in the loop?

aspacca: you cannot mix http response body and http response headers.

all the the headers must be sent before the body.

you can collect in a slice all the `getURL(r, s.proxyPort).ResolveReference(relativeURL).String()` at https://github.com/dutchcoders/transfer.sh/pull/435/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR394

and after you sent all the header, range over the slice and execute `write fmt.Fprintln(w, etc`

aspacca:
@anihm136 are you still going to work on this?

otherwise I will create a new branch with the fix, thanks :)

anihm136:
Sorry, I'm a little packed right now. Please go ahead with the fix

ISPmng:
merge pull requests

anihm136:
Hey @aspacca I have some time on my hands for the next week or so. Shall I go ahead and finish this?

aspacca:
@anihm136 thanks

could you fix the conflict with main branch?

there was some refactoring since you opened your branch

anihm136:
Should be done

aspacca (approved):

aspacca:
@anihm136 can you fix the linting?

thanks

anihm136:
Done. Unfortunately I missed it because my editor uses the `--fast` flag with the linter, which missed the error

server/handlers.go
```diff
@@ -384,9 +386,17 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 
 			filename = url.PathEscape(filename)
 			relativeURL, _ := url.Parse(path.Join(s.proxyPath, token, filename))
-			_, _ = w.Write([]byte(getURL(r, s.proxyPort).ResolveReference(relativeURL).String()))
+			deleteURL, _ := url.Parse(path.Join(s.proxyPath, token, filename, metadata.DeletionToken))
+			w.Header().Add("X-Url-Delete", resolveURL(r, deleteURL, s.proxyPort))
+			responseBody += fmt.Sprintln( getURL(r, s.proxyPort).ResolveReference(relativeURL).String())
+			s.cleanTmpFile(file)
```
stefanbenten, line 392: We should not need this call here, due to the defer in L319.

stefanbenten (changes requested):

PR #441, opened 2021-12-07
fix in force-https redirect

### LINKED ISSUE #440: Endless redirects using SSL/TLS with docker-image
Hi!

I'm trying to run the dutchcoders/transfer.sh:latest docker image, utilizing the TLS/SSL feature.

The closest I've come, using the host-os's cert and priv-keys:

```bash
docker run --publish 8080:8080 --publish 443:443 \
	--volume /etc/letsencrypt:/etc/letsencrypt dutchcoders/transfer.sh:latest \
	--provider local \
	--basedir /tmp/ \
	--force-https true \
	--tls-cert-file /etc/letsencrypt/live/my.domain.com/fullchain.pem \
	--tls-private-key /etc/letsencrypt/live/my.domain.com/privkey.pem \
	--tls-listener :443
```

But I'm faced with endless 308 redirects when doing this. 
Is this because the Dockerfile only exposes 8080 I presume?

**Apache proxy workaround**
As a temporary workaround I'm able to make the frondend https using a Apache2 proxy setup, but download-links and all the URL's are still set to http://, thus most browser are blocking any downloads I'm sharing directly.

```apache2
<VirtualHost *:443>

    ProxyPreserveHost On
    ProxyRequests Off
    ServerName my.domain.com
    ServerAlias my.domain.com
    ProxyPass / http://localhost:8080/
    ProxyPassReverse / http://localhost:8080/


    ErrorLog ${APACHE_LOG_DIR}/gule.cloud-error.log
    CustomLog ${APACHE_LOG_DIR}/gule.cloud-access.log combined

SSLCertificateFile /etc/letsencrypt/live/my.domain.com/fullchain.pem
SSLCertificateKeyFile /etc/letsencrypt/live/my.domain.com/privkey.pem
Include /etc/letsencrypt/options-ssl-apache.conf
</VirtualHost>

```

**Suggestions**
* Would it be possible to add a hostname/domain-flag (--domain https:://my.domain.com) or other prefix to the parameters?
* I've also managed to hack a half-assed workaround using --proxy-path https://my.domain.com, but this screws up all the base-url's in the examples and the frontend at large.

francescayeye:
fixes #440

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -630,25 +630,24 @@ func getURL(r *http.Request, proxyPort string) *url.URL {
 		u.Scheme = "http"
 	}
 
-	if u.Host == "" {
-		host, port, err := net.SplitHostPort(r.Host)
-		if err != nil {
-			host = r.Host
-			port = ""
-		}
-		if len(proxyPort) != 0 {
-			port = proxyPort
-		}
-		if len(port) == 0 {
+	host, port, err := net.SplitHostPort(r.Host)
+	if err != nil {
+		host = r.Host
+		port = ""
+	}
+	if len(proxyPort) != 0 {
+		port = proxyPort
+	}
+
+	if len(port) == 0 {
+		u.Host = host
+	} else {
+		if port == "80" && u.Scheme == "http" {
+			u.Host = host
+		} else if port == "443" && u.Scheme == "https" {
 			u.Host = host
 		} else {
-			if port == "80" && u.Scheme == "http" {
-				u.Host = host
-			} else if port == "443" && u.Scheme == "https" {
-				u.Host = host
-			} else {
-				u.Host = net.JoinHostPort(host, port)
-			}
+			u.Host = net.JoinHostPort(host, port)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


server/handlers.go
```diff
@@ -1099,10 +1098,22 @@ func (s *Server) RedirectHandler(h http.Handler) http.HandlerFunc {
 		} else if strings.HasSuffix(ipAddrFromRemoteAddr(r.Host), ".onion") {
 			// .onion addresses cannot get a valid certificate, so don't redirect
 		} else if r.Header.Get("X-Forwarded-Proto") == "https" {
-		} else if r.URL.Scheme == "https" {
+		} else if r.TLS != nil {
 		} else {
 			u := getURL(r, s.proxyPort)
 			u.Scheme = "https"
+			if len(s.proxyPort) == 0 && len(s.TLSListenerString) > 0 {
+				_, port, err := net.SplitHostPort(s.TLSListenerString)
```
stefanbenten, line 1106: Should we assign host here and use it further down the same way as we do in the first block of code above?

aspacca: indeed not. in the first block of code we use the `request.Host`
here we use the `TLSListenerString`, that's in the format `x.x.x.x:port` or even only `:port`

we should use only the port of the tls listener, and only in the case a proxy port is not set, and `u.Hostname()` for the host (that will be the one coming from the first block/request)

stefanbenten: True. In my case i had it setup to actually have the hostname:port configuration, but this is certainly more flexible.

stefanbenten (approved):

PR #452, opened 2021-12-25
server/handlers.go,storage.go: smaller fixes

stefanbenten:
This change moves a couple of defers in order to prevent them from not being executed since we have returned in error cases early.
It also fixes a couple of smaller linting issues, like error messages not starting with a capital letter and variable name collisions with imported packages.

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -367,12 +367,12 @@ func (s *Server) postHandler(w http.ResponseWriter, r *http.Request) {
 			buffer := &bytes.Buffer{}
 			if err := json.NewEncoder(buffer).Encode(metadata); err != nil {
 				s.logger.Printf("%s", err.Error())
-				http.Error(w, errors.New("could not encode metadata").Error(), 500)
+				http.Error(w, "Could not encode metadata", 500)
 
 				return
 			} else if err := s.storage.Put(token, fmt.Sprintf("%s.metadata", filename), buffer, "text/json", uint64(buffer.Len())); err != nil {
 				s.logger.Printf("%s", err.Error())
-				http.Error(w, errors.New("could not save metadata").Error(), 500)
+				http.Error(w, "Could not save metadata", 500)
 
 				return
 			}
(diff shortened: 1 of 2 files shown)


server/handlers.go
```diff
@@ -94,7 +94,7 @@ func initHTMLTemplates() *html_template.Template {
 }
 
 func healthHandler(w http.ResponseWriter, r *http.Request) {
-	fmt.Fprintf(w, "Approaching Neutral Zone, all systems normal and functioning.")
+	_, _ = w.Write([]byte("Approaching Neutral Zone, all systems normal and functioning."))
```
aspacca, line 97: what' the ration behind this change?

stefanbenten: According to the docs i found it is recommended to directly use the Write Method. I am ok changing this back tho.

aspacca: I don't have a strong opinione, but seems that the previous doesn't need error handling/ignoring

I would rather replace with `fmt.Fprintln`, unless we actually have any print formatting

stefanbenten: fmt.Fprintln actually required error handling and ignoring, just was omitted until now

aspacca: I should have checked on https://pkg.go.dev/fmt#Fprintf :D 

then it's fine, let's keep `w.Write`

omitting `_, _ = ` maybe?

aspacca:
another consideration:
while `errors.New()` should not start with a capital letter, I so no particolar reason why we wrap the string for `http.Error()` in `errors.New()` and extract its string value with `Error()` 

being a message in the http body I see it proper to start with a capital letter, and we could pass the message directly as a string. what do you think?

server/handlers.go
stefanbenten: I do plan on hooking up something like `golangci-lint` soon to the repo and at latest then we would need to add it :D

stefanbenten:
I agree, we do not have to create the errors in those lines. Just wanted to make the linter i am running locally more "happy".

aspacca:
do we need to merge https://github.com/dutchcoders/transfer.sh/pull/453 before, and then apply `CloseCheck` here as well?

stefanbenten:
Yeah lets do that.

aspacca (approved):

PR #453, opened 2021-12-25
.github/workflows: adding golangci-lint as new job

stefanbenten:
In order to follow golang's best practices, we should lint the code base properly beyond usual syntax mistakes.


server/handlers.go
```diff
@@ -755,11 +748,9 @@ func (s *Server) purgeHandler() {
...
 	go func() {
 		for {
-			select {
-			case <-ticker.C:
-				err := s.storage.Purge(s.purgeDays)
-				s.logger.Printf("error cleaning up expired files: %v", err)
-			}
+			<-ticker.C
+			err := s.storage.Purge(s.purgeDays)
+			s.logger.Printf("error cleaning up expired files: %v", err)
```
aspacca, line 753: let's add an `err != nil` in order to log, what do you think?

server/storage.go
```diff
@@ -337,7 +337,7 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
 		return nil, err
 	}
 
-	srv, err := drive.New(getGDriveClient(config, localConfigPath, logger))
+	srv, err := drive.New(getGDriveClient(config, localConfigPath, logger)) // nolint: staticcheck
```
aspacca, line 340: why this is needed?

server/handlers.go
stefanbenten: Done

server/storage.go
stefanbenten: It is needed because the drive.New function is deprecated and would need some work to update it.
Should i add a ToDo on top?

aspacca:
good, can I merge?

aspacca (approved):

stefanbenten:
Yes, good to merge!

PR #468, opened 2022-02-17
.github/,go.mod,go.sum: Update storj dependencies, drop older go versions

stefanbenten:
- Bumps storj dep to v1.8.2
- Drops go v1.13 and go v1.14 since they are outdated and not supported by some recent dependencies
- Bumps go version for testing to latest stable release v1.18


aspacca:
@stefanbenten can you look at 1.13 and 1.14 failure? we could also remove them in case

stefanbenten:
I'll look into it. Seems like an unwanted dependency bump in the uplink package.

stefanbenten:
@aspacca Since Go v1.18 is now released (and soon available for github actions), i dropped older v1.13 and v1.14 and prepared the workflows to use the latest version.

stefanbenten:
@aspacca This should be good to go now.

francescayeye (approved):

PR #474, opened 2022-03-14
Employ ACME Handler for HTTP01-Challenge

### LINKED ISSUE #473: Go 1.17 strict TLS APLN breaks Let's Encrypt
More information and a fix noted here: https://github.com/traefik/traefik/issues/8376

Logs
```
transfer    | 2022/03/14 15:26:18 http: TLS handshake error from 18.192.36.99:37146: tls: client requested unsupported application protocols ([acme-tls/1])
transfer    | 2022/03/14 15:26:18 http: TLS handshake error from 34.221.255.206:22562: tls: client requested unsupported application protocols ([acme-tls/1])
transfer    | 2022/03/14 15:26:18 http: TLS handshake error from 18.116.86.117:55338: tls: client requested unsupported application protocols ([acme-tls/1])
transfer    | 2022/03/14 15:26:18 http: TLS handshake error from 64.78.149.164:29096: tls: client requested unsupported application protocols ([acme-tls/1])
transfer    | 2022/03/14 15:26:19 http: TLS handshake error from xx.xx.xx.xx:39743: acme/autocert: unable to satisfy "https://acme-v02.api.letsencrypt.org/acme/authz-v3/9999999999" for domain "xxxx.xxx": no viable challenge type found
transfer    | 2022/03/14 15:26:19 http: TLS handshake error from xx.xx.xx.xx:39751: acme/autocert: missing certificate
```

docweirdo:
Fixes #473 by employing the HTTPHandler provided by `acme.autocert.Manager`. 

I am still not sure why the error `tls: client requested unsupported application protocols ([acme-tls/1])` is thrown when trying to obtain a certificate via ACME. The parameter `nextProtos` of the TLS config is not set and the source code shows the error is only thrown when this parameter is not empty :shrug:

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -28,7 +28,6 @@ import (
 	crypto_rand "crypto/rand"
 	"encoding/binary"
 	"errors"
-	gorillaHandlers "github.com/gorilla/handlers"
 	"log"
 	"math/rand"
 	"mime"
@@ -40,6 +39,8 @@ import (
 	"sync"
 	"syscall"
 	"time"
+
+	gorillaHandlers "github.com/gorilla/handlers"
 
 	context "golang.org/x/net/context"
 
@@ -56,8 +57,9 @@ import (
 	web "github.com/dutchcoders/transfer.sh-web"
 	assetfs "github.com/elazarl/go-bindata-assetfs"
 
-	autocert "golang.org/x/crypto/acme/autocert"
 	"path/filepath"
+
+	autocert "golang.org/x/crypto/acme/autocert"
 )
 
 // parse request with maximum memory of _24Kilobits
@@ -259,7 +261,7 @@ func UseLetsEncrypt(hosts []string) OptionFn {
 	return func(srvr *Server) {
 		cacheDir := "./cache/"
 
-		m := autocert.Manager{
+		srvr.autocertManager = &autocert.Manager{
 			Prompt: autocert.AcceptTOS,
 			Cache:  autocert.DirCache(cacheDir),
 			HostPolicy: func(_ context.Context, host string) error {
@@ -278,7 +280,7 @@ func UseLetsEncrypt(hosts []string) OptionFn {
 		}
 
 		srvr.tlsConfig = &tls.Config{
-			GetCertificate: m.GetCertificate,
+			GetCertificate: srvr.autocertManager.GetCertificate,
 		}
 	}
 }
@@ -325,7 +327,8 @@ type Server struct {
 
 	logger *log.Logger
 
-	tlsConfig *tls.Config
+	tlsConfig       *tls.Config
+	autocertManager *autocert.Manager
 
 	profilerEnabled bool
 
@@ -532,7 +535,8 @@ func (s *Server) Run() {
 		ipFilterHandler(
 			handlers.LogHandler(
 				LoveHandler(
-					s.RedirectHandler(cors(r))),
+					s.autocertManager.HTTPHandler(
+						s.RedirectHandler(cors(r)))),
 				handlers.NewLogOptions(s.logger.Printf, "_default_"),
 			),
 			s.ipFilterOptions,


aspacca:
@docweirdo thanks for the fix

indeed I'm not sure this won't have any impact in case direct tls config is used, instead of let's encrypt

I can read the following here:
https://cs.opensource.google/go/x/crypto/+/master:acme/autocert/autocert.go;l=242-243
```
// If GetCertificate is used directly, instead of via Manager.TLSConfig, package users will
// also have to add acme.ALPNProto to NextProtos for tls-alpn-01, or use HTTPHandler for http-01.
```

Using `acme.Manager.TLSConfig()` does indeed add `acme.ALPNProto` to `nextProtos`:
https://cs.opensource.google/go/x/crypto/+/master:acme/autocert/autocert.go;l=225-227

so maybe @BenHarris was on the right path

do you mind trying to create another PR mimicking the `nextProtos` as setup by `acme.Manager.TLSConfig()` and check if it will solve the issue as well?

If so I would go for that solution, since it will be isolated to when let's encrypt is enabled

BenHarris:
For what it's worth, adding `NextProtos: []string{"h2", "http/1.1", "acme-tls/1"},` to `srvr.tlsConfig = &tls.Config{` in `server/server.go` seems to sort it and is contained within the UseLetsEncrypt function.

docweirdo:
Ok then I guess my solution is unnecessary. I will play around with your suggestions hopefully today or tomorrow.

I still don't really understand why this problem arises or more precisely, where `nextProtos` is set currently to throw the observes errors, or is the Documentation just not accurate?

BenHarris:
Just looked into the autocert source, and @aspacca's solution probably looks to be better. Ultimately both approaches lead to the same outcome but it probably makes sense to use the autocert implementation.

aspacca:
@docweirdo 

> I still don't really understand why this problem arises or more precisely, where nextProtos is set currently to throw the observes errors, or is the Documentation just not accurate?

I could not find the documentation you are reffering to

but here's where the error from @BenHarris is returned: https://github.com/golang/go/blob/7dd10d4ce20e64d96a10cb67794851a58d96a2aa/src/crypto/tls/handshake_server.go#L303

I either `serverProtos` or `clientProtos` is empty it will return before the error:
https://github.com/golang/go/blob/7dd10d4ce20e64d96a10cb67794851a58d96a2aa/src/crypto/tls/handshake_server.go#L282-L284

`clientProtos` seems to be set here:
https://github.com/golang/go/blob/cc5e3de593afca73cf1b4d732ddceffb2837b390/src/crypto/tls/handshake_messages.go#L501

I'm not sure where `serverProtos` is set, maybe somewhere else in the autocert package

aspacca:
> This is what confuses me, as `tlsConfig` is currently empty except for `getCertificate`.

https://github.com/golang/go/blob/go1.17.8/src/net/http/server.go#L3058-L3062

empty slice for `config.NextProtos` -> if condition returns true -> `config.NextProtos` is then not empty, and we jump here: https://github.com/golang/go/blob/7dd10d4ce20e64d96a10cb67794851a58d96a2aa/src/crypto/tls/handshake_server.go#L286

> I also do not quite understand how using `Manager.TLSConfig` (or just setting `nextProtos` manually like @BenHarris did it) enables any handling of a HTTP challenge on port 80. Could it be that this solution only allows for the TLS-ALPN-01 challenge?

you are right. `autocert.Manager.HTTPHandler` intercepts requests to `/.well-known/acme-challenge/` to provision HTTP-01 challenge responses. but it is not required since letsencrypt supports TLS-ALPN-01 challenge as well, that is fulfilled by `autocert.Manager.GetCertificate`

I would stick to the current implementation and just add @BenHarris proposed fix, since HTTP-01 requires a plain listener on 80, that could be disable in transfer.sh by `tls-listener-only` flag

docweirdo:
Thanks for the links. Sounds good to me.

PR #486, opened 2022-05-14
Issue 485

### LINKED ISSUE #485: "Hostname" and "webAddress" printed in punycode instead of unicode (IDN)
Hey there, 
in order to use this software with intertionalized domain names (IDN) is auspicable to print unicode instead of punycode.
 
you can see a live example here -> https://краш.мкд

francescayeye:
closes #485


server/handlers.go
```diff
@@ -36,6 +36,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"golang.org/x/net/idna"
```
stefanbenten, line 39: Lets move this down into the right section.

stefanbenten (commented):
Small linting/style suggestion, otherwise looks good to me.

server/handlers.go
aspacca: hoped it was spotted by the linter :)

stefanbenten (approved):

server/handlers.go
stefanbenten: I'll add a lint step for that 👍

PR #488, opened 2022-05-21
Fix for unrecognized flags on du

frankievalentine:
--total --block-size="K" --dereference were all unrecognized flags for this script. I have updated them to the correct flags to run correctly.


stefanbenten (approved):
Works fine for me on Linux and Mac 👍

Issue #490, opened 2022-06-02
How can I resume downloads?


niss88:
I'm downloading from transfer.sh url like wget  -c "https://transfer.sh/..." but in the middle way an error occurs stating that "...バイトで読み込みエラーが発生しました (Operation timed out)。 再試行しています。". 
Then it restarts downloading but from the scratch (progress bar starts from 0%). I read this https://github.com/dutchcoders/transfer.sh/issues/289 but didn't get the solution.

aspacca:
from wget documentation
```
-c
       --continue
           Continue getting a partially-downloaded file.  This is useful when you want to finish up a download started by a previous instance of Wget, or by another program.  For instance:

                   wget -c ftp://sunsite.doc.ic.ac.uk/ls-lR.Z

           If there is a file named ls-lR.Z in the current directory, Wget will assume that it is the first portion of the remote file, and will ask the server to continue the retrieval from an
           offset equal to the length of the local file.

           Note that you don't need to specify this option if you just want the current invocation of Wget to retry downloading a file should the connection be lost midway through.  This is the
           default behavior.  -c only affects resumption of downloads started prior to this invocation of Wget, and whose local files are still sitting around.

           Without -c, the previous example would just download the remote file to ls-lR.Z.1, leaving the truncated ls-lR.Z file alone.
```

niss88:
Sorry but I'm always putting "wget -c url" and what's wrong with that

aspacca:
I'm sorry @niss88, you are right
range donwload works only for video and audio file: https://github.com/dutchcoders/transfer.sh/blob/main/server/handlers.go#L1054

I will fix as soon as I have time

niss88:
My downloading file was a video... but I'll wait either way.

For testing I uploaded mp4 file for example https://transfer.sh/KnDl1c/out3.mp4 (~1 MB), tested like below and --continue (-c) didn't work for mp4 file
```
*****.. [13:56:54] tmp% ls                                         
*****.. [13:57:40] tmp% wget -c https://transfer.sh/KnDl1c/out3.mp4
--2022-06-03 13:57:55--  https://transfer.sh/KnDl1c/out3.mp4
transfer.sh (transfer.sh) をDNSに問いあわせています... 144.76.136.153, 2a01:4f8:200:1097::2
transfer.sh (transfer.sh)|144.76.136.153|:443 に接続しています... 接続しました。
HTTP による接続要求を送信しました、応答を待っています... 200 OK
[...]

Issue #409, opened 2021-09-14
Bug: Error 404 when try to delete file uploaded via zsh function


salmannotkhan:
First of it doesn't give any deletion token when we upload file via zsh function/plugin
Another thing is it shows delete button whenever someone visits the link and it prompts for deletion token but when we click on confirm button it shows "Error (404) during deletion of file"

So, I'll suggest to remove delete button completely. and also I want to know when files deleted from the server?

aspacca:
We won't remove the delete button, since it's working properly: you have to provide the deletion token, as stated in the hint.
When you click the button a `DELETE` request is made to `token/filename/deletionToken`, if the deletion token is wrong or empty, the request result in a 404 (not found)

Deletion token is returned on upload in the header `X-Url-Delete`, zsh function doesn't print it: I never used zsh so I wouldn't be able to modify. Any external PR to enanche the zsh function is welcome

aspacca:
> I want to know when files deleted from the server

every server can setup it's own purge period.
If the server you've uploaded to is using `v1.3.0` the purge time is reported in the homepage

salmannotkhan:
So I think it's useless to show that delete button

salmannotkhan:
> Deletion token is returned on upload in the header `X-Url-Delete`, zsh function doesn't print it: I never used zsh so I wouldn't be able to modify. Any external PR to enanche the zsh function is welcome

Okay, I'll try that. Also when you click confirm again and again it appends the error into same error instead of replacing it completely.

salmannotkhan:
Where I can find the script for transfer cli command?

salmannotkhan:
>Also when you click confirm again and again it appends the error into same error instead of replacing it completely.

nvm I made PR on transfer.sh-web repo for that fix

imTHAI:
> Deletion token is returned on upload in the header `X-Url-Delete`, zsh function doesn't print it: I never used zsh so I wouldn't be able to modify. Any external PR to enanche the zsh function is welcome

That's why ! I was searching where the hell I should be supposed to have received it. Maybe it would be good to warm people somewhere in the README.md text, zsh is even the default shell under MacOS

corporateuser:
Some command line clients do not print headers by default, please consider to return JSON object instead of a simple string:
```JSON
{
  downloadUrl: string,
  deletionURL: string
}
```

aspacca:
@corporateuser sorry, that will be a breaking change

PR #495, opened 2022-07-10
allow range request for every content type, move lock only when metad…

francescayeye:
…ata has to be updated

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -686,9 +686,6 @@ func (s *Server) unlock(token, filename string) {
 }
 
 func (s *Server) checkMetadata(ctx context.Context, token, filename string, increaseDownload bool) (metadata, error) {
-	s.lock(token, filename)
-	defer s.unlock(token, filename)
-
 	var metadata metadata
 
 	r, _, err := s.storage.Get(ctx, token, fmt.Sprintf("%s.metadata", filename))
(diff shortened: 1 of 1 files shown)


aspacca:
related to #409 

see https://github.com/dutchcoders/transfer.sh/issues/490#issuecomment-1179545800 and https://github.com/dutchcoders/transfer.sh/issues/490#issuecomment-1179654566  


@stefanbenten can you take a look when you have time?

this is to add the support on `transfer.sh` service, hosted instances like https://transfer.sh might require tuninque eventual proxy on top of the service

server/handlers.go
```diff
@@ -1051,7 +1065,7 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 		reader = ioutil.NopCloser(bluemonday.UGCPolicy().SanitizeReader(reader))
 	}
 
-	if w.Header().Get("Range") != "" || strings.HasPrefix(metadata.ContentType, "video") || strings.HasPrefix(metadata.ContentType, "audio") {
+	if r.Header.Get("Range") != "" {
```
stefanbenten, line 1068: This creates a very big amount of unnecessary downloads of the entire file and thus lots of load on the backend storage (for cases of remote storage via full downloads and for local storage due to copying the file entirely into a temp file) in L1078.

stefanbenten (changes requested):
This needs a lot more work to do proper range requests without having to download the entire file. I have been trying to plumb the range requests through to the backend and let it handle it accordingly. Tho this either requires a change of the backend storage interface or creating lots logic and temporary files from within the storage provider (google storage being the only one tho).

stefanbenten:
Let me see, if i can push this PR sometime today..

server/handlers.go
aspacca: it just creates them for every type of file content instead of video and audio files
where was a bug (`w` vs `r`) in the original implementation that prevented range requests to be fulfilled on non audio/video files

but yes, we should try to exploit range donwload directly from the storage implementation.
we can change the `storage.Get` interface to receive the range to pass down to the concrete storage "sdk" in order to minimise what's actually downloaded

good catch anyway: mine was a quick hack and I didn't think about the performance implications

aspacca:
closed in favor of #527 

we still should keep https://github.com/dutchcoders/transfer.sh/pull/495/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR705-R706, for donwloaders that send multiple concurrent range request

PR #501, opened 2022-08-11
Fixed improper implementation of content type

blind-intruder:
Add text/plain content type if metadata.ContentType is unable to determine the content type of the file

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -1038,7 +1038,7 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 		metadata.ContentType is unable to determine the type of the content, 
 		So add text/plain in this case to fix XSS related issues/
 		*/
-                if(contentType == ""){
+		if strings.TrimSpace(contentType) == "" {
 		      contentType := "text/plain"
 		 }
 	} else {
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -1039,7 +1039,7 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 		So add text/plain in this case to fix XSS related issues/
 		*/
 		if strings.TrimSpace(contentType) == "" {
-		      contentType := "text/plain"
+		      contentType = "text/plain"
 		 }
 	} else {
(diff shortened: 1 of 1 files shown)


server/handlers.go
```diff
@@ -1034,6 +1034,11 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 
 	if action == "inline" {
 		disposition = "inline"
+		/*
```
aspacca, line 1037: the comment does not seem correct :)

also, I would expect https://github.com/dutchcoders/transfer.sh/pull/501/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR1056-R1058 to sanitise the content and avoid XSS

if the above is not working it should be fixed there (or on the bluemonday module)

aspacca (commented):
see my comment @blind-intruder , thanks :)

server/handlers.go
```diff
@@ -1034,6 +1034,11 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
...
+		metadata.ContentType is unable to determine the type of the content, 
+		So add text/plain in this case to fix XSS related issues/
+		*/
+		contentType := "text/plain"
```
aspacca, line 1041: let's override `contentType` only if it's empty, and in case let's fix what's missing in https://github.com/dutchcoders/transfer.sh/blob/64c7759126ffea425c55d76dd49cbbc1a6719992/server/handlers.go#L97-L116

thanks

aspacca (commented):
I mislooked some datails, please see my updated comment

server/handlers.go
aspacca: or maybe:
https://github.com/dutchcoders/transfer.sh/blob/main/server/handlers.go#L1051
```
if (disposition == "inline" && canContainsXSS(contentType)) || (contentType == "") {
```


@stefanbenten what do you thini it's the best approach?

aspacca:
@blind-intruder , I took the liberty to fix for the linter and add trimming

ready to merge as soon as CI will be green

PR #512, opened 2022-10-25
GDrive: Fix fieldNotWritable, Add SupportSharedDrive, Add service_account method

DoyunShin:


server/storage/gdrive.go
```diff
@@ -84,7 +92,7 @@ func (s *GDrive) setupRoot() error {
 		MimeType: gDriveDirectoryMimeType,
 	}
 
-	di, err := s.service.Files.Create(dir).Fields("id").Do()
+	di, err := s.service.Files.Create(dir).Fields("id").SupportsAllDrives(true).Do()
```
aspacca, line 95: that's magic! thanks :)

DoyunShin: oh forgot to change.. hehe

DoyunShin:
Didnt expectd gDriveRootConfigFile requires localConfigPath on commit cd9c56e96237721fd295cea848ca80aab5a5a385
reverted with 73ac766cd7486ae5c452383e7961d71da1350a56

server/storage/gdrive.go
```diff
@@ -45,13 +46,20 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
 	}
 
 	// If modifying these scopes, delete your previously saved client_secret.json.
-	config, err := google.ConfigFromJSON(b, drive.DriveScope, drive.DriveMetadataScope)
-	if err != nil {
-		return nil, err
+	var httpClient *http.Client
+
+	if strings.Contains(string(b), `"type": "service_account"`) {
```
aspacca, line 51: let's have a separate cli param for service account file

server/storage/gdrive.go
```diff
@@ -45,13 +46,20 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
...
+		logger.Println("GDrive: using Service Account credentials")
+		httpClient = JWTConfigFromJSON(b, logger).Client(ctx)
+	} else {
+		logger.Println("GDrive: using OAuth2 credentials")
+		config, err := google.ConfigFromJSON(b, drive.DriveScope, drive.DriveMetadataScope)
+		if err != nil {
+			return nil, err
+		}
+		httpClient = getGDriveClient(ctx, config, localConfigPath, logger)
```
aspacca, line 60: let's rename `getGDriveClient` to something like `getGDriveClientFromToken`

server/storage/gdrive.go
```diff
@@ -45,13 +46,20 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
...
 
 	// If modifying these scopes, delete your previously saved client_secret.json.
-	config, err := google.ConfigFromJSON(b, drive.DriveScope, drive.DriveMetadataScope)
-	if err != nil {
-		return nil, err
+	var httpClient *http.Client
+
+	if strings.Contains(string(b), `"type": "service_account"`) {
+		logger.Println("GDrive: using Service Account credentials")
+		httpClient = JWTConfigFromJSON(b, logger).Client(ctx)
```
aspacca, line 53: `JWTConfigFromJSON` -> `getGDriveClientFromServiceAccount`

no need to pass `logger`, it's not used in the function

aspacca:
how do you select a shared drive as target?
currently we write the `root` of transfer.sh in the main folder, and save the id
https://github.com/dutchcoders/transfer.sh/blob/main/server/storage/gdrive.go#L87-L96

is giving the service account access to a specific shared folder that you end up using a folder? (total guess)

DoyunShin:
> is giving the service account access to a specific shared folder that you end up using a folder? (total guess)

image
Yep. By giving a permission to service account(name@projectname.iam.gserviceaccount.com), the service account can access to shared folder

Actually service account doesnt have root folder, so i think we can get folder id from basedir
(and if basedir is "root" then get root folder)

server/storage/gdrive.go
DoyunShin: param like --gdrive-type=service / oauth2?
or we can actually check in json.

DoyunShin: Oauth2
`
{
    "installed": {
        "client_id": "clientid",
        "project_id": "projectname",
        "auth_uri": "https://accounts.google.com/o/oauth2/auth",
        "token_uri": "https://oauth2.googleapis.com/token",
        "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
        "client_secret": "clientsecret"
    }
}
`

Service Account
`
{
  "type": "service_account",
  "project_id": "projectname",
  "private_key_id": "keyid",
  "private_key": "-----BEGIN PRIVATE KEY-----\nKEYSECRET\n-----END PRIVATE KEY-----\n",
  "client_email": "transfersh@projectname.iam.gserviceaccount.com",
  "client_id": "clientid",
  "auth_uri": "https://accounts.google.com/o/oauth2/auth",
  "token_uri": "https://oauth2.googleapis.com/token",
  "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
  "client_x509_cert_url": "https://www.googleapis.com/robot/v1/metadata/x509/transfersh%40projectname.iam.gserviceaccount.com"
}
`

server/storage/gdrive.go
DoyunShin: Done with fb87a5b32f8acda48d5a7f8ec582ce4f5bdee8c8

server/storage/gdrive.go
DoyunShin: Done with f0963001c3e330dd2c3ef0cefe796382c58a6cb4

DoyunShin:
be00bb8afa1aeed07c3a7fdf75d3b3d362402ef6
Changed to Using basedir as rootID.

if basedir == "root", then user's drive root folder is used to store files.

DoyunShin:
https://github.com/dutchcoders/transfer.sh/pull/512#discussion_r1005139290

About this,
https://github.com/dutchcoders/transfer.sh/pull/512#discussion_r1005157344
This is the credentials of service account and oauth2.

There is no problem even if we receive the auth-type with cli or automatically identify it with credentials.json.
Which choice do we go?

Or, it seems that there will be no problem even if it is made to work both when auth-type is entered or not.

server/storage/gdrive.go
aspacca: > param like --gdrive-type=service / oauth2?
> or we can actually check in json.

param like `gdrive-client-json-filepath`: `gdrive-service-account-json-filepath`

only one of the two can be set

in the factory you can translate this to a single param with the path and a new param for the type (service account/oauth2): it is up to you

server/storage/gdrive.go
```diff
@@ -45,65 +44,66 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
...
-	di, err := s.service.Files.Create(dir).Fields("id").Do()
-	if err != nil {
-		return err
-	}
-
-	s.rootID = di.Id
-	err = ioutil.WriteFile(rootFileConfig, []byte(s.rootID), os.FileMode(0600))
-	if err != nil {
-		return err
+func (s *GDrive) checkRoot() error {
```
aspacca, line 88: this is a breaking change.

basedir is the name of the folder to be created in the gdrive, rootID is the id of the folder once created and that we save. the two carry two different information.

if we replce rootID value with basedir value:
1. we don't create the folder
2. current installations have to change the value that they pass to basedir, to use instead rootID

I guess you meant this change to be the way to select a shared drive, is it?

server/storage/gdrive.go
```diff
@@ -337,10 +336,13 @@ func getGDriveClient(ctx context.Context, config *oauth2.Config, localConfigPath
...
 // Request a token from the web, then returns the retrieved token.
 func getGDriveTokenFromWeb(ctx context.Context, config *oauth2.Config, logger *log.Logger) *oauth2.Token {
+	config.RedirectURL = "urn:ietf:wg:oauth:2.0:oob"
 	authURL := config.AuthCodeURL("state-token", oauth2.AccessTypeOffline)
+
 	fmt.Printf("Go to the following link in your browser then type the "+
-		"authorization code: \n%v\n", authURL)
+		"authorization code.\n%v\n", authURL)
 
+	fmt.Printf("Authorization code: ")
```
aspacca, line 345: 👍

server/storage/gdrive.go
```diff
@@ -45,65 +44,66 @@ func NewGDriveStorage(clientJSONFilepath string, localConfigPath string, basedir
...
+	return err
 }
 
 func (s *GDrive) hasChecksum(f *drive.File) bool {
 	return f.Md5Checksum != ""
 }
 
 func (s *GDrive) list(nextPageToken string, q string) (*drive.FileList, error) {
-	return s.service.Files.List().Fields("nextPageToken, files(id, name, mimeType)").Q(q).PageToken(nextPageToken).Do()
+	return s.service.Files.List().Fields("nextPageToken, files(id, name, mimeType)").Q(q).PageToken(nextPageToken).SupportsAllDrives(true).IncludeItemsFromAllDrives(true).Do()
```
aspacca, line 106: why here you need `IncludeItemsFromAllDrives()` and not somewhere else?

DoyunShin: IncludeItemsFromAllDrives() is only need to List()
(checked on gdrive go docs)

server/storage/gdrive.go
DoyunShin: Yep!

I changed basedir to insert folder id, to select rootID.

DoyunShin:
> in the factory you can translate this to a single param with the path and a new param for the type (service account/oauth2): it is up to you

Added gdrive-auth-type
oauth2|service_account

server/storage/gdrive.go
aspacca: I cannot accept a breaking change, sorry
you have to find another way to provide the sharedfolder to use _and_ the basedir as it was

cmd/cmd.go
```diff
@@ -169,6 +169,12 @@ var globalFlags = []cli.Flag{
 		Value:  "",
 		EnvVar: "GDRIVE_CLIENT_JSON_FILEPATH",
 	},
+	cli.StringFlag{
+		Name:   "gdrive-auth-type",
```
aspacca, line 173: this way you pass `gdrive-client-json-filepath` when it's a `gdrive-service-account-filepath` indeed

also it's a breaking change, I cannot accept it, sorry :(

you can add `gdrive-service-account-filepath`, check that only this or `gdrive-client-json-filepath` is set (and at least one of them), and set the `authType` value to pass to the factory: https://github.com/dutchcoders/transfer.sh/pull/512/files#diff-8e494a434a8037b6c0b888e25b2baae7618fe65e792d4a155dadd096e9350667R488

please, remember to change the name of the param in the factory to something generic, like `gdriveCredentialsFilepath`

aspacca (changes requested):
please. see my latest comments :)

Issue #522, opened 2023-01-04
--listener should allow specifying the network interface too, not just the port


mpl:
As is customary with such an option for a server, when the interface is not specified, such as in `--listener :8080`, then we should listen on all interfaces, but otherwise it should be taken into account.

For example: `--listener localhost:8080`, or `--listener 192.168.0.12:8080`, etc.

mpl:
Nevermind, I was fooled by your log line that says "listening on port:" (it should say "listening on addr" or something like that instead), but it actually works since you're passing the option straight as the Addr of the Server.

Sorry about the noise.

aspacca:
> Nevermind, I was fooled by your log line that says "listening on port:" (it should say "listening on addr"

I will change the log entry :)

mpl:
@aspacca cool.

Btw, on that subject, I would advise you not to ignore the error at 

https://github.com/dutchcoders/transfer.sh/blob/main/server/server.go#L549

because you will silently miss errors such as when the port is busy, or we're not allowed to listen on the interface, etc.
For example, on my machine, 
`
	log.Fatal(http.ListenAndServe("localhost:777", nil))
`
would result in 
`
2023/01/05 10:53:19 listen tcp 127.0.0.1:777: bind: permission denied

But transfer.sh will be happily silent about it:
`
[transfer.sh]2023/01/05 10:45:06 listening on port: localhost:777
[transfer.sh]2023/01/05 10:45:06 ---------------------------
`
And the user will never know that the listening actually failed.

aspacca:
good catch! thanks

btw, feel free to open a PR for it if you have time :)

mpl:
sure, thanks. I might do that then ;)

PR #523, opened 2023-01-05
server: do not ignore listening errors

mpl:
Also fixes the logging message so it does not make believe that --listen is only about the port.
And tiny refactors to make the HTTP server and HTTPS servers setup similar, to ease maintenance.

Context: #522


aspacca:
great @mpl , thanks!

PR #536, opened 2023-03-12
Add `Vary` headers in responses

kotx:
This prevents caching servers like Cloudflare from caching responses when `Accept` or `X-Decrypt-Password` request headers vary. Before this, servers like Cloudflare would cache the html preview page regardless of `Accept` header (meaning curls and non-browser requests would also return a cached preview page). Adding a `Vary` header should prevent that.

More info here: https://blog.cloudflare.com/vary-for-images-serve-the-correct-images-to-the-correct-browsers/

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -245,7 +245,7 @@ func canContainsXSS(contentType string) bool {
 
 /* The preview handler will show a preview of the content for browsers (accept type text/html), and referer is not transfer.sh */
 func (s *Server) previewHandler(w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Vary", "Accept, Referer, X-Decrypt-Password")
+	w.Header().Set("Vary", "Range, Referer, X-Decrypt-Password")
 
 	vars := mux.Vars(r)
 
@@ -1160,7 +1160,7 @@ func (s *Server) headHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Connection", "close")
 	w.Header().Set("X-Remaining-Downloads", remainingDownloads)
 	w.Header().Set("X-Remaining-Days", remainingDays)
-	w.Header().Set("Vary", "Accept, Referer, X-Decrypt-Password")
+	w.Header().Set("Vary", "Range, Referer, X-Decrypt-Password")
 
 	if s.storage.IsRangeSupported() {
 		w.Header().Set("Accept-Ranges", "bytes")
@@ -1254,7 +1254,7 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", contentType)
 	w.Header().Set("Content-Length", strconv.FormatUint(contentLength, 10))
-	w.Header().Set("Vary", "Accept, Referer, X-Decrypt-Password")
+	w.Header().Set("Vary", "Range, Referer, X-Decrypt-Password")
 
 	if _, err = io.Copy(w, decryptionReader); err != nil {
 		s.logger.Printf("%s", err.Error())
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -245,7 +245,7 @@ func canContainsXSS(contentType string) bool {
 
 /* The preview handler will show a preview of the content for browsers (accept type text/html), and referer is not transfer.sh */
 func (s *Server) previewHandler(w http.ResponseWriter, r *http.Request) {
-	w.Header().Set("Vary", "Accept, X-Decrypt-Password")
+	w.Header().Set("Vary", "Accept, Referer, X-Decrypt-Password")
 
 	vars := mux.Vars(r)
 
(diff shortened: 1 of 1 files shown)


server/handlers.go
```diff
@@ -243,6 +243,8 @@ func canContainsXSS(contentType string) bool {
 
 /* The preview handler will show a preview of the content for browsers (accept type text/html), and referer is not transfer.sh */
 func (s *Server) previewHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Vary", "Accept, X-Decrypt-Password")
```
aspacca, line 246: `previewHandler ` should not be sensible to `X-Decrypt-Password`

aspacca: indeed not event to `Accept`, or do I miss anything?

server/handlers.go
```diff
@@ -1248,6 +1252,7 @@ func (s *Server) getHandler(w http.ResponseWriter, r *http.Request) {
 
 	w.Header().Set("Content-Type", contentType)
 	w.Header().Set("Content-Length", strconv.FormatUint(contentLength, 10))
+	w.Header().Set("Vary", "Accept, X-Decrypt-Password")
```
aspacca, line 1255: unless I've missed something `getHandler` is sensible only to `Range` and `X-Decrypt-Password`

server/handlers.go
```diff
@@ -1155,6 +1158,7 @@ func (s *Server) headHandler(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Connection", "close")
 	w.Header().Set("X-Remaining-Downloads", remainingDownloads)
 	w.Header().Set("X-Remaining-Days", remainingDays)
+	w.Header().Set("Vary", "Accept, X-Decrypt-Password")
```
aspacca, line 1161: unless I've missed something `headHandler` is not sensible to any header

aspacca (commented):
@kotx thank you very much for the contribution

please, see my comments

aspacca:
> More info here: https://blog.cloudflare.com/vary-for-images-serve-the-correct-images-to-the-correct-browsers/

I'm not 100% sure about the cloudflare behaviour on the `Vary` header: the article mentions only `Accept`, is just for the sake of the example?

If not, and they only "react" to `Accept` it seems to me that they are forcing the upstream service behind their cache to send a value that does not respect the real ones that should be set: in this case I would set the proper ones, and if it still does not work with cloudflare cache please ask them to fix ;)

server/handlers.go
kotx: I think the Vary header [should always be the same for the same URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary).

kotx:
> > More info here: https://blog.cloudflare.com/vary-for-images-serve-the-correct-images-to-the-correct-browsers/
> 
> I'm not 100% sure about the cloudflare behaviour on the `Vary` header: the article mentions only `Accept`, is just for the sake of the example?

I've only just discovered this, but apparently Cloudflare only allows [Vary for images](https://developers.cloudflare.com/cache/about/vary-for-images) on the non-free plans? :(

But normally, Cloudflare and any server should not cache if the headers in `Vary` are different.

aspacca:
> But normally, Cloudflare and any server should not cache if the headers in `Vary` are different.

yes, I'm find adding the `Vary` header, just I won't implement it to make cloudflare happy, but according to the rfc :)

server/handlers.go
aspacca: >I think the Vary header [should always be the same for the same URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Vary).

I had this doubt just after leaving the comment

the for `previewHandler`, `getHandler` and `headHandler` we must set `Range, X-Decrypt-Password`
for `viewHandler` only `Accept`

kotx: `viewHandler` is only for `/`, yes? I don't know if it should vary based on `X-Decrypt-Password`, only for the file route.
Is that okay?

aspacca: yes, `viewHandler` is only for `/`, no need to vary based on `X-Decrypt-Password`

indeed the file route must vary also for `Referer`

kotx: Should be fixed in the latest commit

aspacca (approved):

server/handlers.go
aspacca: sorry, file route should have `w.Header().Set("Vary", "Range, Referer, X-Decrypt-Password")`

kotx: Sorry, I thought getHandler was the file route, do you mean zipHandler, tarHandler, tarGzHandler, etc should also have the vary header?

aspacca: these three ones:

https://github.com/dutchcoders/transfer.sh/pull/536/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR246
https://github.com/dutchcoders/transfer.sh/pull/536/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR1161
https://github.com/dutchcoders/transfer.sh/pull/536/files#diff-961509ed3ff668f3172148150903514fb3394d6ff090d6db78d986875b13689fR1255

currently they have `w.Header().Set("Vary", "Accept, Referer, X-Decrypt-Password")`
it should be `w.Header().Set("Vary", "Range, Referer, X-Decrypt-Password")`

response changes based on `Accept` only for `/` (`viewHandler`)
reponse for `/{token}/{filename}` changes `Range, Referer, X-Decrypt-Password`

`Range` and `X-Decrypt-Password` are used on `getHandler`
according to the `Referer` value we route to `previewHandler` or `getHandler`

finally `headHandler` should have the same `Vary` value since this should be the same based on URL and not URL+HTTP Method

kotx: Got it, so sorry for the confusion!
Made the changes you requested

aspacca (approved):

aspacca:
@kotx could you please merge the main branch on yours?
thanks

kotx:
> could you please merge the main branch on yours? thanks

Done, hope this works

Issue #546, opened 2023-04-28
Can't upload files with spaces in filename


mpr1255:
Problem is simple. This works: 

`echo 'hello' > test1.txt && transfer test1.txt`

This doesn't: 

`echo 'hello' > "Test 2.txt" && transfer "Test 2.txt"`

Update: I just asked GPT-4 and it proposed this solution which I trialed and it works well. I also got it to fix it for bash & zsh compatibility. 

The key was just the URL encoding function so instead of passing the filename straight, it encodes it. Neat!

```
transfer () {
    if [ $# -eq 0 ]
    then
        echo "No arguments specified.\nUsage:\n transfer <file|directory>\n ... | transfer <file_name>" >&2
        return 1
    fi
    if tty -s
    then
        file="$1"
        file_name=$(basename "$file")
        if [ ! -e "$file" ]
        then
            echo "$file: No such file or directory" >&2
            return 1
        fi
        if [ -d "$file" ]
        then
            file_name="$file_name.zip"
            (
                cd "$file" && zip -r -q - .
            ) | curl --progress-bar --upload-file "-" "https://transfer.sh/$(urlencode "$file_name")" | tee /dev/null
        else
            cat "$file" | curl --progress-bar --upload-file "-" "https://transfer.sh/$(urlencode "$file_name")" | tee /dev/null
        fi
    else
        file_name=$1
        curl --progress-bar --upload-file "-" "https://transfer.sh/$(urlencode "$file_name")" | tee /dev/null
    fi
}

urlencode() {
    local string="${1}"
    local length="${#string}"
    for (( i=0; i<$length; i++ )); do
        local c="${string:$i:1}"
        case $c in
            [a-zA-Z0-9.~_-]) printf "$c" ;;
            *) printf '%%%02X' "'$c"
        esac
    done
}
```

aspacca:
@mpr1255 this was reported in https://github.com/dutchcoders/transfer.sh/issues/483 and have an alternative shell function at https://github.com/dutchcoders/transfer.sh#bash-and-zsh-with-delete-url-delete-token-output-and-prompt-before-uploading

the `transfer` shell function in the frontend was not fixed.
there's no need to urlencode the filename, not providing a filename and using the file and not stdin for `--upload-file` (without the `cat $file|`) is enough:
```
curl -v --progress-bar --upload-file hello\ 2 "https://transfer.sh/"
```

I would rather go for this solution, since it decreases a lot the overall complexity of the function

PR #548, opened 2023-04-29
Add URL encoding to transfer function to handle whitespaces

snowphone:
Add URL encoding functionality to the `transfer` function in `README.md` so that file names with spaces can be uploaded correctly (e.g. `hello world.json`).

Also, the code uses the `_urlencode` function written in Python3. I know not every *nix OS includes Python3 by default but I'm familiar with Python and I don't know how to make it more portable.

This commit includes changes to the following:

- Update the `transfer` function to use `_urlencode` for encoding the file names as URL.
- Add the `_urlencode` function to URL encode the file names.

Since this oneliner is too hard to review, I uploaded formatted code too:
[before.txt](https://github.com/dutchcoders/transfer.sh/files/11359288/before.txt)
[after.txt](https://github.com/dutchcoders/transfer.sh/files/11359289/after.txt)

If this PR is okay, I'll make a PR to tranfer.sh-web too

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ You need to create an OAuth Client id from console.cloud.google.com, download th
 ### Bash and zsh (multiple files uploaded as zip archive)
 ##### Add this to .bashrc or .zshrc or its equivalent
 ```bash
-transfer(){ if [ $# -eq 0 ];then echo "No arguments specified.\nUsage:\n transfer <file|directory>\n ... | transfer <file_name>">&2;return 1;fi;if tty -s;then file="$1";file_name=$(basename "$file");if [ ! -e "$file" ];then echo "$file: No such file or directory">&2;return 1;fi;if [ -d "$file" ];then file_name="$file_name.zip" ,;(cd "$file"&&zip -r -q - .)|curl --progress-bar --upload-file "-" "https://transfer.sh/$file_name"|tee /dev/null,;else cat "$file"|curl --progress-bar --upload-file "-" "https://transfer.sh/$file_name"|tee /dev/null;fi;else file_name=$1;curl --progress-bar --upload-file "-" "https://transfer.sh/$file_name"|tee /dev/null;fi;}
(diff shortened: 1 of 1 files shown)


aspacca:
hi @snowphone , thanks for the PR

please, see my comment at https://github.com/dutchcoders/transfer.sh/issues/546#issuecomment-1528347712

the best solution would be to avoid `cat` and do not add the filename in the url
```
cat "$file"|curl --progress-bar --upload-file "-" "https://transfer.sh/$file_name"
```
vs
```
curl --progress-bar --upload-file "$file" "https://transfer.sh/"
```

so no need for urlencode and $file_name

snowphone:
Wow it's too simple 😂
I changed this PR as you advised.

README.md
```diff
@@ -243,7 +243,7 @@ You need to create an OAuth Client id from console.cloud.google.com, download th
...
+transfer(){ if [ $# -eq 0 ];then echo "No arguments specified.\nUsage:\n transfer <file|directory>\n ... | transfer <file_name>">&2;return 1;fi;if tty -s;then file="$1";file_name=$(basename "$file");if [ ! -e "$file" ];then echo "$file: No such file or directory">&2;return 1;fi;if [ -d "$file" ];then file_name="$file_name.zip" ,;(cd "$file"&&zip -r -q - .)|curl --progress-bar --upload-file "-" "https://transfer.sh/"|tee /dev/null;else cat "$file"|curl --progress-bar --upload-file "-" "https://transfer.sh/"|tee /dev/null;fi;else file_name=$1;curl --progress-bar --upload-file "-" "https://transfer.sh/"|tee /dev/null;fi;}
```
aspacca, line 246:
```
if [ -d "$file" ];then file_name="$file_name.zip" ,;(cd "$file"&&zip -r -q - .)|curl --progress-bar --upload-file "-" "https://transfer.sh/"|tee /dev/null;
```
here `$file` is a directory, we cd into it and zip recursively sending output to stdout, curl get from stdin the upload file from stdin.
we should change to zip to file and use it for curl


```
else cat "$file"|curl --progress-bar --upload-file "-" "https://transfer.sh/"|tee /dev/null;fi
```

same here: it's not enough to remove `$file_name`, we should remove `cat` and curl getting the upload file from stdin:
```
else curl --progress-bar --upload-file "$file" "https://transfer.sh/"|tee /dev/null;fi
```


```
transfer(){ 
    if [ $# -eq 0 ];then 
        echo "No arguments specified.\nUsage:\n transfer <file|directory>\n ... | transfer <file_name>">&2;
        return 1;
    fi;
    if tty -s; then 
        file_name=$(basename "$1");
        if [ ! -e "$1" ];then 
            echo "$1: No such file or directory">&2;
            return 1;
        fi;
        if [ -d   "$1" ];then 
            zip_file="$1.zip";
            pushd "$1";
            zip -r -q "$zip_file" . && curl --progress-bar --upload-file "$zip_file" "https://transfer.sh/"|tee /dev/null && rm "$zip_file";
            popd;
        else 
            curl --progress-bar --upload-file "$file" "https://transfer.sh/"|tee /dev/null;
        fi;
    else 
        curl --progress-bar --upload-file "$1" "https://transfer.sh/"|tee /dev/null;
    fi;
}
```

this should be the proper one: please @snowphone check if it works for you and feel free to push the change :)

thanks

aspacca (changes requested):

PR #550, opened 2023-05-11
Dockerfile: Use Go 1.20 by default

adamantike:
Updating the default Go version used in the Dockerfile, to the latest one supported by the project at the moment.

Also, Go `1.17` support was removed in https://github.com/dutchcoders/transfer.sh/commit/9f1fe62e05115bd355799be7a633231686363c48


stefanbenten:
Thank you for the contribution @adamantike

PR #551, opened 2023-05-11
Upgrade github.com/urfave/cli to v2

adamantike:
`urfave/cli` v2 was released in 2019, and it doesn't introduce relevant user-facing breaking changes.

Migration guide from v1 can be found at: https://cli.urfave.org/migrate-v1-to-v2/

The only user-facing change is within the "Flags before args" section, but I haven't found that to affect `transfer.sh` as arguments are not used.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"os"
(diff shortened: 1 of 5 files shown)


aspacca (approved):

aspacca:
thanks for this @adamantike

aspacca (approved):

PR #558, opened 2023-05-18
Improve purgeTime display in web page

### LINKED ISSUE #557: purgeTime display
Currently the purge time is displayed with the format hours, minutes, seconds. For example, I want to store the files for 30 days, but in the web page shows "Files stored for 720h0m0s":
Screenshot from 2023-05-17 23-39-47

I have thought of two different approaches:

**1. Display only the days** without the hours and seconds:
Screenshot from 2023-05-17 23-37-52

**2. Display the days, hours and seconds**, omitting zero values:
Screenshot from 2023-05-17 23-35-13

I know that at the moment transfer.sh only supports days (an int value) for the purge time. If this is always going to be the case, perhaps it would be better to only show the days the files will be stored.
But, if there are plans or ideas to use another duration, such as minutes or seconds, in addition to days, the second approach would be better.

I'm willing to work on this, after deciding which approach to take.

natilou:
- changing the line `purgeTime = s.purgeDays.String()` to use a function that formats the days like this: "N days" or "1 day"
- adding the function `formatDurationDays` in utils.go file

Fixes #557

Screenshot from 2023-05-18 16-53-35
Screenshot from 2023-05-17 23-37-52

### DIFF
--- a/server/handlers.go
+++ b/server/handlers.go
@@ -374,7 +374,7 @@ func (s *Server) viewHandler(w http.ResponseWriter, r *http.Request) {
 
 	purgeTime := ""
 	if s.purgeDays > 0 {
-		purgeTime = s.purgeDays.String()
+		purgeTime = formatDurationDays(s.purgeDays)
 	}
 
 	data := struct {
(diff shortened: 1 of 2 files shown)


aspacca (approved):

aspacca:
thanks @natilou !

PR #559, opened 2023-05-22
Upgrade aws-sdk-go to v2

adamantike:
`aws-sdk-go-v2` is the newer SDK version, replacing the one being used at the moment by the project.

This change maintains full compatibility with existing flags and configurations, and only replaces the underlying library.


server/storage/s3.go
```diff
@@ -2,38 +2,61 @@ package storage
...
 	purgeDays   time.Duration
 	noMultipart bool
 }
 
 // NewS3Storage is the factory for S3Storage
-func NewS3Storage(accessKey, secretKey, bucketName string, purgeDays int, region, endpoint string, disableMultipart bool, forcePathStyle bool, logger *log.Logger) (*S3Storage, error) {
-	sess := getAwsSession(accessKey, secretKey, region, endpoint, forcePathStyle)
+func NewS3Storage(ctx context.Context, accessKey, secretKey, bucketName string, purgeDays int, region, endpoint string, disableMultipart bool, forcePathStyle bool, logger *log.Logger) (*S3Storage, error) {
+	cfg, err := config.LoadDefaultConfig(ctx,
+		config.WithEndpointResolverWithOptions(aws.EndpointResolverWithOptionsFunc(
```
aspacca, line 32: I've some doubt here: if `endpoint` is empty (most of the case) the endppoint resolver will return `aws.EndpointNotFoundError`

rather I'd expect to set an `aws.EndpointResolverWithOptionsFunc` only if `endpoint` is not empty, or set it in the s3 service like it is done at https://github.com/elastic/beats/blob/main/x-pack/filebeat/input/awss3/input.go#L238-L242

server/storage/s3.go
```diff
@@ -177,12 +190,3 @@ func (s *S3Storage) Put(ctx context.Context, token string, filename string, read
 }
 
 func (s *S3Storage) IsRangeSupported() bool { return true }
-
-func getAwsSession(accessKey, secretKey, region, endpoint string, forcePathStyle bool) *session.Session {
```
aspacca, line 181: let's keep a separated `getAwsConfig` function, to use at https://github.com/dutchcoders/transfer.sh/pull/559/files#diff-66e5af6447b0120f4ac7cd603a574ed6ca422560f3d03e8012f7a4f6acfb0326R32-R51

aspacca (changes requested):
thanks @adamantike , this was long due! :)

server/storage/s3.go
adamantike: Either way should work fine. My initial implementation checked for `endpoint` length, but then I checked the official example: https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/endpoints/#overriding-endpoint-with-fallback

> An endpoint resolver can use the [EndpointNotFoundError](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#EndpointNotFoundError) sentinel error value to trigger fallback resolution to the service clients default resolution logic. This allows you to selectively override one or more endpoints seamlessly without having to handle fallback logic.

However, as the app is only using S3, I can move this logic to only configure the S3 client instead.

adamantike:
PR updated based on feedback!

aspacca (approved):

PR #605, opened 2024-03-06
Generate secure token using crypto rand

rumanzo:
rand package https://pkg.go.dev/math/rand notice us
`
Package rand implements pseudo-random number generators suitable for tasks such as simulation, but it should not be used for security-sensitive work.
`
I purpose use crypto/rand function that's more secure

### DIFF
--- a/server/token.go
+++ b/server/token.go
@@ -25,7 +25,9 @@ THE SOFTWARE.
 package server
 
 import (
-	"math/rand"
+	"crypto/rand"
+	"log"
+	"math/big"
 )
 
 const (
@@ -37,8 +39,11 @@ const (
 func token(length int) string {
 	result := ""
 	for i := 0; i < length; i++ {
-		x := rand.Intn(len(SYMBOLS) - 1)
-		result = string(SYMBOLS[x]) + result
+		x, err := rand.Int(rand.Reader, big.NewInt(int64(len(SYMBOLS))))
+		if err != nil {
+			log.Fatal("Failed to generate token")
+		}
+		result = string(SYMBOLS[x.Int64()]) + result
 	}
 
 	return result


server/token.go
```diff
@@ -37,8 +39,11 @@ const (
 func token(length int) string {
 	result := ""
 	for i := 0; i < length; i++ {
-		x := rand.Intn(len(SYMBOLS) - 1)
-		result = string(SYMBOLS[x]) + result
+		x, err := rand.Int(rand.Reader, big.NewInt(int64(len(SYMBOLS))))
+		if err != nil {
+			log.Fatal("Failed to generate token")
```
aspacca, line 44: we should return an error and gracefully fail the request with a reason to the user

aspacca (changes requested):

aspacca:
hello @rumanzo , thanks for the PR, please see my comment

server/token.go
rumanzo: Can we just ignore err in this case?
I really doubt that we can get error in this case ever using rand.Reader
https://cs.opensource.google/go/go/+/master:src/crypto/rand/util.go;l=85
And we defenetly won't  panic there

aspacca: digging further, that's on unix system:
https://cs.opensource.google/go/go/+/master:src/crypto/rand/rand_unix.go;l=62
https://cs.opensource.google/go/go/+/master:src/crypto/rand/rand_unix.go;l=72

It's not excluded 100% that they won't return an error.

Indeed I guess that on our docker image (that's stratch based) the first will always return an error: do you mind trying to build the docker image from your branch and test?

thanks!

aspacca (commented):
please, see my comment

server/token.go
rumanzo: `
Indeed I guess that on our docker image (that's stratch based) the first will always return an error: do you mind trying to build the docker image from your branch and test?
`
I tried run scratch image with dropped all capabilites, and it works. But i can do something like fallback

aspacca: I can reproduce and this seems to be the reason:
```
// On Linux, FreeBSD, Dragonfly, NetBSD and Solaris, Reader uses getrandom(2) if
// available, /dev/urandom otherwise.
```

but, please, note:
```
// On other Unix-like systems, Reader reads from /dev/urandom.
```

so I guess on other Unix-like systems it won't work.

not sure how to test

rumanzo: `I can reproduce and this seems to be the reason:`
how?

`so I guess on other Unix-like systems it won't work.`
on other unix-like system should be /dev/urandom, but what another unix-like system? HP-UX may be?

rumanzo:
please see my comment in code review

server/token.go
aspacca: > `I can reproduce and this seems to be the reason:` how?
> 
> `so I guess on other Unix-like systems it won't work.` on other unix-like system should be /dev/urandom, but what another unix-like system? HP-UX may be?

let's check the error and panic with a message with the link to this PR

rumanzo: > > `I can reproduce and this seems to be the reason:` how?
> > `so I guess on other Unix-like systems it won't work.` on other unix-like system should be /dev/urandom, but what another unix-like system? HP-UX may be?
> 
> let's check the error and panic with a message with the link to this PR

What you mean? How?

server/token.go
```diff
@@ -35,11 +37,14 @@ const (
 
 // generate a token
 func token(length int) string {
-	result := ""
+	result := make([]byte, length)
 	for i := 0; i < length; i++ {
-		x := rand.Intn(len(SYMBOLS) - 1)
-		result = string(SYMBOLS[x]) + result
+		x, err := rand.Int(rand.Reader, big.NewInt(int64(len(SYMBOLS))))
+		if err != nil {
```
aspacca, line 43:
```suggestion
		if err != nil {
		// log an error entry and fallback to `math/rand`
```

server/token.go
aspacca: > What you mean? How?

nevermind!

see my last suggestion please :)

server/token.go
rumanzo: I agree, will do it later. Should I pass logger into function?

aspacca: yes, thanks

you should pass a null logger in the test: https://github.com/dutchcoders/transfer.sh/blob/main/server/token_test.go

rumanzo: I did it. But error message in logger in this function provide three error messages in log, because token calls several times.

Issue #505, opened 2022-08-30
Update version of go to >=1.18?


joshspicer:
For personal use, I currently working on adding support for Azure blob storage as an alternative storage driver (https://github.com/dutchcoders/transfer.sh/compare/main...joshspicer:transfer.sh:main?expand=1).  

I would love to contribute back the Azure storage driver back to this repo. The Azure SDKs require at least go 1.18.  I am curious if there's any justification for pinning this project to go 1.15, which is now unsupported.

stefanbenten:
Usually its good to require the latest + the previous to minor versions (ie 1.19, 1.18, 1.17) and thus setting the minimum to 1.17 would be a good intermediate step imo.
For Azure i think there is a "deprecated" sdk that can be used that builds with 1.16 and upwards?

aspacca:
> Usually its good to require the latest + the previous two minor versions (ie 1.19, 1.18, 1.17)

in order to clarify: my goal is to give users the possibility to build from source on any LTS (or equivalent) major system they use, relying on the default repo of the system package manager (or equivalent).

MacOS: hombrew is almost up to latest release, same macports - no problem here
Windows: I honestly don't have a clue :)
Linux:
- Debian, buster, from backports: 1.15
- Anything that can fallback on EPEL, from EPEL7: 1.17

aspacca:
to shorten the discussion:  @joshspicer you should be able to pin to a non-latest version of the azure sdk and still use azure blob storage

is it ok for you?

joshspicer:
FWIW the official stance from the golang team is to [support two major versions](https://go.dev/doc/devel/release) back from the latest (so 1.17, 1.18, and 1.19 today).

I'll see if I can get my changes working with 1.17 so that things are backwards compatible.  I would prefer to use the new Azure Blob Storage sdk (https://github.com/Azure/azure-sdk-for-go).  Not sure if all the features of this SDK require 1.18, or if I can somehow get away with using the SDK in 1.17

aspacca:
as far as I know github.com/Azure/azure-sdk-for-go@v59.0.0+incompatible works with 1.17

aspacca:
@joshspicer are this still an issue for you?

Did you add support for azure blob storage?

stefanbenten:
@aspacca Debian bullseye is the latest release and its backports support 1.19. I dont think we should support anything older than 1.17 anymore, in the best case i would like even only supporting 1.18+. for the same reasons outlined by Josh above.

[...]

PR #638, opened 2025-01-30
Add Azure blob storage backend

joshspicer:
finally getting around to https://github.com/dutchcoders/transfer.sh/issues/505 😅


server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func getCredentials() (*azidentity.DefaultAzureCredential, error) {
+	return azidentity.NewDefaultAzureCredential(nil)
+}
+
+func NewAzureBlobStorage(ctx context.Context, storageAccountName string, containerName string, logger *log.Logger) (Storage, error) {
+	logger.Println("Creating Azure Blob Storage")
+
+	if storageAccountName == "" {
```
aspacca, line 30: you check for this on https://github.com/dutchcoders/transfer.sh/pull/638/files#diff-8e494a434a8037b6c0b888e25b2baae7618fe65e792d4a155dadd096e9350667R511. please do not repeat the check and leave it only on `cmd.go`

server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+	if containerName == "" {
+		return nil, fmt.Errorf("missing container name")
+	}
+
+	credentials, err := getCredentials()
+	if err != nil {
+		return nil, err
+	}
+
+	serviceUrl := "https://" + storageAccountName + ".blob.core.windows.net"
```
aspacca, line 43: please. use `fmt.Sprintf` :)

server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+	return azureStorage, nil
+}
+
+func (s *AzureStorage) Type() string {
+	return "azure"
+}
+
+func (s *AzureStorage) Get(ctx context.Context, token string, filename string, _ *Range) (io.ReadCloser, uint64, error) {
+	key := fmt.Sprintf("%s/%s", token, filename)
+	blobClient := s.containerClient.NewBlobClient(key)
```
aspacca, line 66: does azure blob storage support content range?
see: https://github.com/dutchcoders/transfer.sh/blob/main/server/storage/s3.go#L103-L105

server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+	key := fmt.Sprintf("%s/%s", token, filename)
+	_, err := s.client.UploadStream(ctx, s.containerName, key, reader, nil)
+	return err
+}
+
+func (s *AzureStorage) Delete(ctx context.Context, token string, filename string) error {
+	key := fmt.Sprintf("%s/%s", token, filename)
+	blobClient := s.containerClient.NewBlobClient(key)
+	_, err := blobClient.Delete(ctx, nil)
+	if err != nil {
```
aspacca, line 93: please, let's keep logging coherent with the other storage: just return `err` with no logging

server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+		s.logger.Printf("Failed to delete blob %s: %v", key, err)
+		return err
+	}
+	s.logger.Printf("Successfully deleted blob %s", key)
+	return nil
+}
+
+func (s *AzureStorage) IsNotExist(err error) bool {
+	return err != nil
```
aspacca, line 102: are you sure about this?
there's not specific `error`for files that not exist in the storage? and is only error that can be returned a file that doesn't exist?

server/storage/azure.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (s *AzureStorage) Purge(ctx context.Context, days time.Duration) error {
+	s.logger.Printf("Purging blobs older than %v days", days)
```
aspacca, line 106: please, keep consistence with other storage logging

cmd/cmd.go
```diff
@@ -495,6 +507,20 @@ func New() *Cmd {
 			} else {
 				options = append(options, server.UseStorage(store))
 			}
+		case "azure":
+			storageAccountName := c.String("azure-storage-account")
+			containerName := c.String("azure-storage-container")
```
aspacca, line 512: please, keep the "if x = ; {} else {}" structure of all the other blocks

aspacca (changes requested):
hello @joshspicer , thank you very much for the contribution!

much appreciated!

please see my comments and we can probably merge this soon

server/storage/azure.go
joshspicer: hmmm not sure yet.  I found https://github.com/Azure/azure-sdk-for-go/blob/sdk/storage/azblob/v1.6.0/sdk/storage/azblob/internal/generated/zz_constants.go#L511 but not sure if this is accessible

aspacca: that's an internal package, one of these should be probably used instead: https://github.com/Azure/azure-sdk-for-go/blob/90c29ccd6952363d37f7f2c2aa2e8d329ee59015/sdk/storage/azfile/fileerror/error_codes.go#L36-L104

@zmoog, sorry if i invoke you: any idea? :)

joshspicer: Indeed, this looks right!  I saw `BlobNotFound` come through by checking for a non-existent deletion token.

image

cmd/cmd.go
```diff
@@ -495,6 +507,16 @@ func New() *Cmd {
 			} else {
 				options = append(options, server.UseStorage(store))
 			}
+		case "azure":
```
aspacca, line 510: 👍 🥇

server/storage/azure.go
```diff
@@ -0,0 +1,129 @@
...
+	return azidentity.NewDefaultAzureCredential(nil)
+}
+
+func NewAzureBlobStorage(ctx context.Context, storageAccountName string, containerName string, logger *log.Logger) (Storage, error) {
+	credentials, err := getCredentials()
+	if err != nil {
+		return nil, err
+	}
+
+	serviceUrl := fmt.Sprintf("https://%s.blob.core.windows.net", storageAccountName)
```
aspacca, line 34: i wonder if there's any compatible azure blob storage around, like there are for s3, that we could support as well if we build the service url in a non-hardcoded way. is there anything like this?

server/storage/azure.go
```diff
@@ -0,0 +1,129 @@
...
+
+func (s *AzureStorage) Type() string {
+	return "azure"
+}
+
+func (s *AzureStorage) Get(ctx context.Context, token string, filename string, rng *Range) (io.ReadCloser, uint64, error) {
+	key := fmt.Sprintf("%s/%s", token, filename)
+	blobClient := s.containerClient.NewBlobClient(key)
+
+	var options *azblob.DownloadStreamOptions
```
aspacca, line 59: 👍 🥇 

can you please make sure that content range works? you can test uploading a video an moving fast fowarded and skipping back and forth in the video player on the donwload page :)

aspacca (approved):
i approve the PR, please just test the content range and change the service url that's currently hard-coded only if my question has a positive answer :)

send a final comment as feedback and i'm going to merge if nothing else is left!

thank you again for the contribution

server/storage/azure.go
joshspicer: Thanks for the idea of how to test, I wasn't familiar with content range before this.  I'll give it a go soon and share the result 😃

joshspicer: I'm able to see the header with your technique but this then results in an error (and the video does not seek forward)

image

I will take a closer look later.  If you have suggestions on where I should place some breakpoints, let me know :)

aspacca: there's a context canceled that should mean that the request to get the blob is timing out.

my guess is that the problem is that you pass `rng.Limit` as the `azblob.HTTPRange.Count`, that in this case is 0, and this might mess up the request, leading it to time out (instead of a more proper bad request status return).

it can't be that you have to set `azblob.HTTPRange.Count` only if `rng.Limit > 0`, because it is not a pointer, so the zero value is already `0`, but it might something like for storj: https://github.com/dutchcoders/transfer.sh/blob/089ff324e13caaa70942ae4b539833a4de561f4e/server/storage/storj.go#L88-L92

i suggest to give it a try with the same `if` logic and `-1`, if it doesn't work let's look in the azblob documentation if anything is specified. if not, i guess we'd have to file a bug to azure for that.

do you have access to some kind of azure support by chance? if so you might consider asking support (directly or after the test and looking at the docs ;))

aspacca: uhm... where does the "CONTENT LENGTH" in your log come from?

it must be from https://github.com/dutchcoders/transfer.sh/pull/638/files/5d3fddfbd6610f258f441f4d3d835a650bffd5af#diff-01d6ca3cf28f771fa37e3c86cd51f957a80fd4327dfa6b4087f9b5ba4a3aaef9R73, but you also have `err != nil`it seems, that must be from https://github.com/dutchcoders/transfer.sh/pull/638/files/5d3fddfbd6610f258f441f4d3d835a650bffd5af#diff-01d6ca3cf28f771fa37e3c86cd51f957a80fd4327dfa6b4087f9b5ba4a3aaef9R71, so i am a little puzzled

aspacca: unless you are taking it form an `HEAD`request :)

server/storage/azure.go
```diff
@@ -0,0 +1,129 @@
...
+	key := fmt.Sprintf("%s/%s", token, filename)
+	_, err := s.client.UploadStream(ctx, s.containerName, key, reader, nil)
+	return err
+}
+
+func (s *AzureStorage) Delete(ctx context.Context, token string, filename string) error {
+	key := fmt.Sprintf("%s/%s", token, filename)
+	blobClient := s.containerClient.NewBlobClient(key)
+	_, err := blobClient.Delete(ctx, nil)
+	if err != nil {
```
aspacca, line 95: you can just `return err` now :)

server/storage/azure.go
```diff
@@ -0,0 +1,129 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, blob := range page.Segment.BlobItems {
+			if time.Since(*blob.Properties.LastModified) > days {
+				blobClient := s.containerClient.NewBlobClient(*blob.Name)
+				_, err := blobClient.Delete(ctx, nil)
+				if err != nil {
+					continue
```
aspacca, line 119: i guess you want to do `return err` here.
continue doesn't do anything, since there are no extra instructions after the if, and `break`would loose the `err` and `return nil` after the whole loop is over

is there a way to initiate a "general `blobClient`" outside of the loop and just override the `blob.Name`, by chance?

aspacca: > is there a way to initiate a "general `blobClient`" outside of the loop and just override the `blob.Name`, by chance?

https://github.com/Azure/azure-sdk-for-go/blob/90c29ccd6952363d37f7f2c2aa2e8d329ee59015/sdk/storage/azblob/client.go#L114-L116

i guess in general you can get rid of the `AzureStorage.containerClient` and `AzureStorage.blobClient` and use the helpers on that package anywhere.

aspacca (changes requested):
oki, sorry, asking for a little few more changes :)

please check my latest comments 🙏 :

server/storage/azure.go
joshspicer: Thank you for all the feedback.  I will revisit soon, apologies I couldn't get to it this weekend :)

joshspicer: Good idea, made that change to not create clients manually.

server/storage/azure.go
joshspicer: Perhaps I need a larger video, but with 8 minute videos i'm not hitting a breakpoint inside the `if rng != nil {...}` block. That said, the experience is ok

Screen Recording 2025-02-17 at 7 25 42 PM

server/storage/azure.go
joshspicer: I am not aware and didn't see anything in a quick search

server/storage/azure.go
aspacca: > Perhaps I need a larger video, but with 8 minute videos i'm not hitting a breakpoint inside the `if rng != nil {...}` block

and scrubbling works anyway?
can you check your network tab and see if any GET request with a content-range header is done?

let's double check the range issue one last time then i'll review the final code and approve (it seems nothing is left, at the moment)

joshspicer: https://github.com/user-attachments/assets/033afb33-22c0-47fe-b1ac-0b030b3b7afa

Tried from https://github.com/dutchcoders/transfer.sh/pull/638/commits/b8c00d87f30ce28a913c766f3499049cf07a8bd0

joshspicer:
@paolafrancesca Still interested in picking up this change, or no? I have happily been using this backend against my fork FWIW :)

francescayeye:
> @paolafrancesca Still interested in picking up this change, or no? I have happily been using this backend against my fork FWIW :)

@joshspicer I'm afraid that the range seems like working in the videos you have attached because the full video was buffered already locally. the `Range` is totally ignored so it's no possibile that `DownloadStream` is doing anything else that downloading the full blob.

the results might be different on slower connection. can you try with a bigger file, like https://archive.org/details/slfullv0.3?

PR #641, opened 2025-02-17
LISTENER env not working, needs `--listener`

ramonsmits:
I tried overriding the port using envvar `LISTENER` but it didn't work, I tried:

```env
LISTENER=8082
LISTENER=:8082
```

It does work via args:

```yml
command: ["--provider", "local", "--listener", ":8082"]
```


aspacca:
hi @ramonsmits , thanks for spotting out a potential bug, i will try to reproduce as soon as possible and come back to you.

i won't accept this PR, anyway, since we should rather fix the bug, if it's confirmed, and not changing the do cumentation to accomodate for the bug ;) :D

PR #659, opened 2025-10-22
update aws credentials build

redwarn:
This PR enhances the AWS configuration logic by introducing support for the AWS default credential provider chain.
When accessKey and secretKey are empty, allowing it to automatically load credentials from multiple AWS-supported sources such as:

**Environment variables**

- Shared credentials/config files (~/.aws/credentials, ~/.aws/config)
- EC2/ECS instance roles
- EKS Pod Identity

This change improves flexibility and simplifies configuration in environments where credentials are already managed by AWS.

**Changes**

Updated getAwsConfig to use the AWS default credential chain when accessKey or secretKey is not provided.
Retains existing behavior (manual static credentials) when both keys are explicitly specified.


redwarn:
sorry , i updated and In my environment the test passed

francescayeye:
this is a breaking change and it should be handled better in case the access key and the secret key are not provided but the default credential chain will fail

PR #661, opened 2025-11-19
Trim prefixed '/' from proxy path

GochoMugo:
This is a minor change.

| before | after |
| --- | --- |
|  before | after |

Alternative solution would be to mention in the documentation that the proxy path should not be prefixed with a `/`.

Also updated README to fix the missing `v` in the docker image tag. 

FYI: Thanks for transfer.sh 👍🏾

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,8 +300,8 @@ nightly| Scheduled CI build every midnight UTC
 nightly-noroot| Scheduled CI build every midnight UTC using [no root]
 edge| Latest CI build after every commit on `main`
 edge-noroot| Latest CI build after every commit on `main` using [no root]
-x.y.z| CI build after tagging a release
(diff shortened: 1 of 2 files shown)


francescayeye:
@GochoMugo , thanks for the PR, but please let's also add in the README that a `/` prefix in the value of the setting will be trimmed. 🙏

GochoMugo:
> @GochoMugo , thanks for the PR, but please let's also add in the README that a `/` prefix in the value of the setting will be trimmed. 🙏

@paolafrancesca Addressed in commit https://github.com/dutchcoders/transfer.sh/pull/661/commits/ce3da8e370223374389a31c554aa3ea30b3579ce.

francescayeye (approved):

PR #667, opened 2026-02-18
Add Helm chart for Kubernetes deployment

Adri3nParra:
Hello, I built this for my work since the project only offers Docker-based deployment today with no official Helm chart.

The chart supports all 4 storage providers (local, S3/S3-compatible, Storj, Google Drive) and comes with secure defaults (non-root, read-only filesystem, capabilities dropped). Ingress, Gateway API, HPA, NetworkPolicy, HTTP auth, and ClamAV integration are all optional and independently configurable.

- Local storage with PVC: tested, working
- S3 with MinIO: tested, working
- Storj: not tested — no access to a Storj account
- Google Drive: not tested — requires OAuth credentials I don't have available

Help testing Storj and Google Drive would be welcome.

### DIFF
--- a/k8s/transfer.sh/templates/configmap.yaml
+++ b/k8s/transfer.sh/templates/configmap.yaml
@@ -1,0 +1,100 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "transfer.sh.fullname" . }}
+  labels:
+    {{- include "transfer.sh.labels" . | nindent 4 }}
+data:
+  PROVIDER: {{ .Values.transfersh.provider | quote }}
+  {{- if .Values.transfersh.purgeDays }}
+  PURGE_DAYS: {{ .Values.transfersh.purgeDays | quote }}
+  {{- end }}
+  {{- if .Values.transfersh.purgeInterval }}
+  PURGE_INTERVAL: {{ .Values.transfersh.purgeInterval | quote }}
+  {{- end }}
+  {{- if .Values.transfersh.maxUploadSize }}
+  MAX_UPLOAD_SIZE: {{ int .Values.transfersh.maxUploadSize | quote }}
+  {{- end }}
+  {{- if .Values.transfersh.randomTokenLength }}
+  RANDOM_TOKEN_LENGTH: {{ .Values.transfersh.randomTokenLength | quote }}
+  {{- end }}
+  {{- /* Local provider */}}
+  {{- if eq .Values.transfersh.provider "local" }}
+  BASEDIR: {{ .Values.transfersh.local.basedir | quote }}
+  {{- end }}
+  {{- /* S3 provider */}}
+  {{- if eq .Values.transfersh.provider "s3" }}
+  {{- with .Values.transfersh.s3 }}
+  BUCKET: {{ .bucket | quote }}
+  S3_REGION: {{ .region | quote }}
+  {{- if .endpoint }}
+  S3_ENDPOINT: {{ .endpoint | quote }}
+  {{- end }}
+  {{- if .pathStyle }}
+  S3_PATH_STYLE: "true"
+  {{- end }}
+  {{- if .noMultipart }}
+  S3_NO_MULTIPART: "true"
+  {{- end }}
+  {{- if and (not .existingSecret) .accessKey }}
+  AWS_ACCESS_KEY: {{ .accessKey | quote }}
+  {{- end }}
+  {{- if and (not .existingSecret) .secretKey }}
+  AWS_SECRET_KEY: {{ .secretKey | quote }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- /* Storj provider */}}
+  {{- if eq .Values.transfersh.provider "storj" }}
+  {{- with .Values.transfersh.storj }}
+  STORJ_BUCKET: {{ .bucket | quote }}
+  {{- if and (not .existingSecret) .access }}
+  STORJ_ACCESS: {{ .access | quote }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- /* Google Drive provider */}}
+  {{- if eq .Values.transfersh.provider "gdrive" }}
+  {{- with .Values.transfersh.gdrive }}
+  BASEDIR: {{ .basedir | quote }}
+  {{- if .clientJsonFilepath }}
+  GDRIVE_CLIENT_JSON_FILEPATH: {{ .clientJsonFilepath | quote }}
+  {{- end }}
+  {{- if .localConfigPath }}
+  GDRIVE_LOCAL_CONFIG_PATH: {{ .localConfigPath | quote }}
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- /* HTTP Authentication */}}
+  {{- if .Values.transfersh.httpAuth.enabled }}
+  {{- if .Values.transfersh.httpAuth.htpasswd }}
+  HTTP_AUTH_HTPASSWD: {{ .Values.transfersh.httpAuth.htpasswd | quote }}
+  {{- else if not .Values.transfersh.httpAuth.existingSecret }}
+  HTTP_AUTH_USER: {{ .Values.transfersh.httpAuth.user | quote }}
+  HTTP_AUTH_PASS: {{ .Values.transfersh.httpAuth.pass | quote }}
+  {{- end }}
+  {{- if .Values.transfersh.httpAuth.ipWhitelist }}
+  HTTP_AUTH_IP_WHITELIST: {{ .Values.transfersh.httpAuth.ipWhitelist | quote }}
+  {{- end }}
+  {{- end }}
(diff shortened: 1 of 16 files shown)


francescayeye:
hello @Adri3nParra 

thank you very much for the contribution: unfortunately I've no possibility to test your helm chart or maintain it, on top of reviewing your PR. 
I won't be able to merge your PR for this reason, since I don't want to end up in the same situation as with the PR for the NIX Flake (#424).

:(

stefanbenten:
@paolafrancesca I am happy to maintain and test this over the weekend 👍

Adri3nParra:
If you encounter any issues during your tests, don’t hesitate to let me know, I can also take a look on my side to help you.

k8s/transfer.sh/templates/configmap.yaml
```diff
@@ -0,0 +1,100 @@
...
+  S3_PATH_STYLE: "true"
+  {{- end }}
+  {{- if .noMultipart }}
+  S3_NO_MULTIPART: "true"
+  {{- end }}
+  {{- if and (not .existingSecret) .accessKey }}
+  AWS_ACCESS_KEY: {{ .accessKey | quote }}
+  {{- end }}
+  {{- if and (not .existingSecret) .secretKey }}
+  AWS_SECRET_KEY: {{ .secretKey | quote }}
```
stefanbenten: Would be great to make them chart-managed secrets instead of plaintext storing them and have the deployment always use the secretKeyRef accordingly.

k8s/transfer.sh/templates/configmap.yaml
```diff
@@ -0,0 +1,100 @@
...
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- /* Storj provider */}}
+  {{- if eq .Values.transfersh.provider "storj" }}
+  {{- with .Values.transfersh.storj }}
+  STORJ_BUCKET: {{ .bucket | quote }}
+  {{- if and (not .existingSecret) .access }}
+  STORJ_ACCESS: {{ .access | quote }}
```
stefanbenten: Same as above

k8s/transfer.sh/templates/configmap.yaml
```diff
@@ -0,0 +1,100 @@
...
+  {{- end }}
+  {{- end }}
+  {{- end }}
+  {{- /* HTTP Authentication */}}
+  {{- if .Values.transfersh.httpAuth.enabled }}
+  {{- if .Values.transfersh.httpAuth.htpasswd }}
+  HTTP_AUTH_HTPASSWD: {{ .Values.transfersh.httpAuth.htpasswd | quote }}
+  {{- else if not .Values.transfersh.httpAuth.existingSecret }}
+  HTTP_AUTH_USER: {{ .Values.transfersh.httpAuth.user | quote }}
+  HTTP_AUTH_PASS: {{ .Values.transfersh.httpAuth.pass | quote }}
```
stefanbenten: Same as above

k8s/transfer.sh/templates/deployment.yaml
```diff
@@ -0,0 +1,130 @@
...
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: 8080
+              protocol: TCP
+          envFrom:
+            - configMapRef:
+                name: {{ include "transfer.sh.fullname" . }}
+          {{- if or (and (eq .Values.transfersh.provider "s3") .Values.transfersh.s3.existingSecret) (and (eq .Values.transfersh.provider "storj") .Values.transfersh.storj.existingSecret) (and .Values.transfersh.httpAuth.enabled .Values.transfersh.httpAuth.existingSecret) }}
```
stefanbenten: This would be good to cover the inline secret management, as referred to in the config map.

k8s/transfer.sh/values.yaml
```diff
@@ -0,0 +1,222 @@
...
+strategy: {}
+
+image:
+  repository: dutchcoders/transfer.sh
+  pullPolicy: IfNotPresent
+  tag: "latest-noroot"
+
+imagePullSecrets: []
+nameOverride: ""
+fullnameOverride: "transfersh"
```
stefanbenten: Would be good to have it default to "", otherwise to deployments in the same namespace would collide.

k8s/transfer.sh/values.yaml
```diff
@@ -0,0 +1,222 @@
...
+    bucket: ""
+    # Use an existing secret containing STORJ_ACCESS key
+    existingSecret: ""
+    # Or set access grant directly
+    access: ""
+
+  # -- Google Drive storage settings (provider: gdrive)
+  gdrive:
+    # Base directory for gdrive storage
+    basedir: ""
```
stefanbenten: Likely want to set something non empty here, otherwise the mountPath should end up invalid.

k8s/transfer.sh/values.yaml
```diff
@@ -0,0 +1,222 @@
...
+    # Path to the OAuth client credentials JSON file (auto-set when using existingSecret)
+    clientJsonFilepath: "/config/gdrive/client.json"
+    # Local config cache directory
+    localConfigPath: "/config/gdrive"
```
stefanbenten: The binary is expecting to write the login token.json into that directory and this seems read-only as per config further down.

stefanbenten (commented):
Sorry for the delayed review @Adri3nParra !
Some smaller security suggestions, otherwise looks sane to me!

k8s/transfer.sh/templates/configmap.yaml
Adri3nParra: Done, moved cred out of CM. For plain text values, chart renders it own secret (see: templates/secret.yaml)

k8s/transfer.sh/templates/configmap.yaml
Adri3nParra: Done, same as above

k8s/transfer.sh/templates/configmap.yaml
Adri3nParra: Done, same as first one
I also found that basicAuthHandler skips the check when user, pass and htpasswd are all empty, so httpAuth.enabled: true with empty values left uploads open. The chart now refuses to render in that case.

k8s/transfer.sh/templates/deployment.yaml
Adri3nParra: Good, env block uses secretKeyRef with "existingSecret | default fullname" as secret name.
Also i add annotation "checksum/config" and "checksum/secret" so when cred are change pods will be rolls.
See : https://helm.sh/fr/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

k8s/transfer.sh/values.yaml
Adri3nParra: Set fullnameOverride to ""

k8s/transfer.sh/values.yaml
Adri3nParra: Fix : chart mounts the data volume at localConfigPath, so it can write token.json and root_id.conf with a RO FS.
Move client.json to /config/client.json.
Feature :  added an optional gdrive.tokenFromSecret. It mounts a token.json key from the gdrive Secret, and the pod starts without waiting on the OAuth prompt. You generate the token once on your computer.

Adri3nParra:
Hello !

I addresses each comment, to sum main fix : 

* S3, Storj and HTTP auth credentials set as plain values go to a chart-managed Secret, and the Deployment reads them through secretKeyRef
* fullnameOverride defaults to ""
* gdrive: basedir holds the Drive folder name, and the data volume sits at localConfigPath so transfer.sh can write there

Minor change : 

* Optional gdrive.tokenFromSecret to mount token.json from the Secret and skip the interactive OAuth flow
* checksum/config and checksum/secret pod annotations, so config and credential changes roll the pods
* The chart refuses to render when httpAuth.enabled has no usable credentials, because the server turns auth off in that case
* tcpSocket liveness/readiness probes: the IP filter covers probe requests too, and kubelet got a 403 from the node IP whenever security.ipWhitelist was set

All is validated with lint and template command from helm on local

k8s/transfer.sh/values.yaml
Adri3nParra: True, now defaults to "transfer.sh", and required check the mount path

stefanbenten:
@Adri3nParra All your commits are unverified/unsigned. Could you kindly check them?

Adri3nParra:
Done, all commits are signed now. Bit of insomnia, it's 4am here :)

stefanbenten (approved):
Thank you for the PR and effort! @Adri3nParra

Adri3nParra:
Thanks for merging @stefanbenten

Would you be interested in a follow-up PR that publishes the chart to GHCR on each GitHub release ?
On your side it would mean making the package public after the first run, and cutting a release to ship it.

